PR #28, opened 2015-09-04
added travel feedrate setting

quillford:
this is useful for faster non cutting/engraving moves


t-oster:
I am not sure which lasercutters will accept feedrate for G0 commands, but I think it won't hurt. The only thing is that it might bloat the job-size since then every alternating G0 and G1 command will contain a feedrate, but that may be not an issue

quillford:
It is only added to the GenericGcodeDriver. I know Marlin and Smoothie accept it for 3D printing. It seems to work exceptionally well on Marlin.

PR #33, opened 2015-11-25
Export gcode

mrzl:


mrzl:
I think this should be mainly it. Will have a look at this tomorrow and recap. In case you got any thoughts let me know :)

src/com/t_oster/liblasercut/drivers/GenericGcodeDriver.java
```diff
@@ -651,6 +659,45 @@ public void sendJob(LaserJob job, ProgressListener pl, List<String> warnings) th
...
+@Override
+public void saveJob(LaserJob job) throws IllegalJobException, Exception {
+	checkJob(job);
+
+	String timestamp = new SimpleDateFormat( "yyyyMMddhhmmssSSS" ).format( new Date( ) );
+	String filename = "output_" + timestamp + ".gcode";
+	System.out.println("Creating file " + filename);
+	this.out = new PrintStream(new File(filename));
+
+	setWaitForOKafterEachLine( false );
```
t-oster: You should save the current value instead of just setting it to false and afterwards to true. What if it was false initially?

src/com/t_oster/liblasercut/drivers/GenericGcodeDriver.java
```diff
@@ -651,6 +659,45 @@ public void sendJob(LaserJob job, ProgressListener pl, List<String> warnings) th
...
   }
+
+@Override
+public void saveJob(LaserJob job) throws IllegalJobException, Exception {
+	checkJob(job);
+
+	String timestamp = new SimpleDateFormat( "yyyyMMddhhmmssSSS" ).format( new Date( ) );
+	String filename = "output_" + timestamp + ".gcode";
+	System.out.println("Creating file " + filename);
+	this.out = new PrintStream(new File(filename));
```
t-oster: We cannot assume, that the current directory is writeable. I think there is some getTemporaryFilename mehtod in IO or somehting like that

t-oster: But in my opinion, the saveJob Mehthod should get an outputStream as a parameter. VisiCut thould then ask the user for a file, create an ouptut stream for that file and give this to the driver.

src/com/t_oster/liblasercut/LaserCutter.java
```diff
@@ -80,6 +80,8 @@ public void sendJob(LaserJob job, ProgressListener pl) throws IllegalJobExceptio
      */
     public abstract void sendJob(LaserJob job, ProgressListener pl, List<String> warnings) throws IllegalJobException, Exception;
 
+	public abstract void saveJob(LaserJob job) throws IllegalJobException, Exception;
```
t-oster: By making this abstract, every LaserCutter has to implement that feature. So either we have to implement it in all drivers now, or at least make this method throw an exception when not overriden.

t-oster:
Hi, thanks for your effort. This is a big step forward.

PR #37, opened 2016-02-12
Fix power command missing on subsequent gcode exports after initial export

madleech:
Using File > Export Laser Gcode gives this on the first run (20mm red circle example on a Grbl based cutter):

```
G21
G90
M3 S0
G0 X21.539200 Y11.226800 F3600
G1 X21.488400 Y12.242800 S1.000000 F50
G1 X21.336000 Y13.258800
```

However if you immediately export it again without changing the power, you get this:

```
G21
G90
M3 S0
G0 X21.539200 Y11.226800 F3600
G1 X21.488400 Y12.242800 F50
G1 X21.336000 Y13.258800
```

Note the missing `S1.000000` command, since `currentPower` hasn't changed from the previous run.


mgmax:
I wonder if it would be a nicer solution to clear currentPower after each M3 S0 command (and also after output of the start gcode), so that we track the device's internal state.

t-oster:
thanks. Looks as if we should refactor sendJob and saveJob to share a common method, because they are 75% the same

PR #38, opened 2016-02-12
Turn laser off when doing G0 rapids

madleech:
Currently the laser is kepy on during G0 rapids.

E.g. two 20mm rectangles next to each other on a grbl:

```
G21
G90
M3 S0
; move to first box
G0 X2.997200 Y2.997200 F3600
; cut first box
G1 X22.961600 Y2.997200 S1.000000 F50
G1 X22.961600 Y22.961600
G1 X2.997200 Y22.961600
G1 X2.997200 Y2.997200
G1 X2.997200 Y2.997200
; move to second box
G0 X29.972000 Y2.997200 F3600
; cut second box
G1 X49.987200 Y2.997200 F50
G1 X49.987200 Y22.961600
G1 X29.972000 Y22.961600
G1 X29.972000 Y2.997200
G1 X29.972000 Y2.997200
M5
G0 X0 Y0
```

The laser is kept on while moving around. Not really an issue unless you have a really powerful laser or are cutting paper.

This commit changes it to turn off the laser during moves:

```
G21
G90
M3 S0
; move to first box
G0 X2.997200 Y2.997200 S0 F3600
; cut first box
G1 X22.961600 Y2.997200 S1.000000 F50
G1 X22.961600 Y22.961600
G1 X2.997200 Y22.961600
G1 X2.997200 Y2.997200
G1 X2.997200 Y2.997200
; move to second box
G0 X29.972000 Y2.997200 S0 F3600
; cut second box
G1 X49.987200 Y2.997200 S1.000000 F50
G1 X49.987200 Y22.961600
G1 X29.972000 Y22.961600
G1 X29.972000 Y2.997200
G1 X29.972000 Y2.997200
M5
G0 X0 Y0
```

Does this seem correct? Or is will it cause issues with non-laser-mode firmwares (e.g. stock grbl) that assume the spindle has inertia and requires a dwell to get up to speed?

### DIFF
--- a/src/com/t_oster/liblasercut/drivers/GenericGcodeDriver.java
+++ b/src/com/t_oster/liblasercut/drivers/GenericGcodeDriver.java
@@ -362,7 +362,8 @@ public class GenericGcodeDriver extends LaserCutter {
     x = isFlipXaxis() ? getBedWidth() - Util.px2mm(x, resolution) : Util.px2mm(x, resolution);
     y = isFlipYaxis() ? getBedHeight() - Util.px2mm(y, resolution) : Util.px2mm(y, resolution);
     currentSpeed = getTravel_speed();
-    sendLine("G0 X%f Y%f F%d", x, y, (int) (travel_speed));
+    currentPower = 0.0;
+    sendLine("G0 X%f Y%f S0 F%d", x, y, (int) (travel_speed));
   }
 
   protected void line(PrintStream out, double x, double y, double resolution) throws IOException {


t-oster:
Well, in my Opinion G0 means "move with laser off", so it should be fixed in the firmware.
However if it does not harm the behavior of other devices, we can merge it.

mgmax:
No, the classical definition of G0 is "uncoordinated move as fast as possible". The (rotating) tool will not be changed. The old DIN standard that defines GCode didn't know about lasers, but G0 never had any side effects like switching things on or off. When milling, you just leave the spindle on until the end of the program which switches it off with M3.

t-oster:
Well then i think there is no reason to not merge this,although i think The smoothie Firmware has laser off on g0

madleech:
Interesting, Smoothieware seems to have its own system (http://smoothieware.org/laser-cutter-guide#toc15), kind of makes sense really. But I'm using MrBeam which is more or less standard Grbl with laser spindle support, and there is no support in Grbl itself for turning off the spindle during rapids, which is why I put it here.

I'll tweak this a bit more so that the Smoothie variant doesn't issue S0 commands on rapids, while Grbl does. @pelrun would using `S0.01` fix tickle power? Or is it better to not issue `S` changes at all for Smoothieboards?

madleech:
... or is this better as a laser cutting settings item, e.g. "Turn off laser during rapids", defaulting to on, except on Smoothie's where it isn't an option and is forced to off?

pelrun:
No, any Sx parameter on a G0 is a full override.

The correct place for this really is in the MrBeam firmware; adding an implied S0 when a G0 is seen.

pelrun:
It even seems like it'd be a trivial change; adding "gc_block.values.s = 0;" to https://github.com/mrbeam/grbl/blob/master/gcode.c#L207 - that'd default to S0 unless there's an explicit Sx on the line.

mgmax:
I find the SmoothieBoard dialect funny because it is a syntax error to combine G0 with an F setting on most classical Gcode interpreters, but they give G0 X... F... as example in their manual.

madleech:
Thanks @pelrun, that does seem pretty simple to change.

From what I can tell:
* Smoothie turns off laser during G0 rapids, however adding Sx parameters alter the behaviour.
* Marlin laser firmware seems to follow the same rules, turning off the laser during G0 rapids, not sure if it does fancy tickle power tricks or not, but probably best to follow the same behaviour as Smoothie firmware.
* Standard Grbl 0.9j doesn't alter laser power during G0 rapids, and so maybe should get `S0` appended during rapids? That said, standard Grbl halts momentarily during every `Sx` power change, and this is the main change that the various laser firmwares make, removing any pauses between power changes. So...
** With the change, Grbl turns off the laser but pauses momentarily before and after each rapid.
** Without the change, Grbl keeps the laser on, but is slightly less fluid.
* MrBeam (based on Grbl 0.9g) also doesn't change laser power during rapids, but could be altered to.

@t-oster do you have an opinion on how this should be handled?
* I patch my firmware to follow the Smoothie standard and this change gets reversed. But, stock Grbl will continue to leave the laser on during G0 rapids.
* I change SmoothieBoard.java and Marlin.java to output G0 rapids with no `Sx` parameter, as they don't need it.
* Or, since I have a few other tweaks in the pipeline for Grbl, such as automatically homing when required and issuing a reset command when first connecting; perhaps I should create a new Grbl driver that inherits from the GenericGcodeDriver, since these changes will be of no interest to Marlin and Smoothie users.

pelrun:
Since you're aiming to support a specific device/class of devices, making a driver for them is a better idea than modifying every other driver to do the reverse.

Do people actually use stock grbl for lasercutter control?

mgmax:
The best solution would be to make it configurable in the generic laser driver and provide suitable default values in the subclasses.

t-oster:
Exactly

madleech:
Thanks for the input everyone, #39 created to address this.

Issue #43, opened 2016-02-22
Provide explicit gcode driver upload method configuration


pelrun:
Currently the gcode driver implicitly chooses an upload type (telnet/http/serial, and I just implemented file in #42) depending on which fields are set in the lasercutter config, which is non-obvious. It'd be much better to have an explicit menu choice.

madleech:
That would be good. Currently it can take a little head scratching to, particularly since both url and host are already set to default values when making a new laser cutter.

PR #48, opened 2016-03-17
Add explicit upload method selection to gcode driver

pelrun:
For #43, adds a upload method selector in the gcode driver options. For pre-existing configurations without the field it will determine it using the original priority logic, so it should automatically be set appropriately. This also provides better error messages if the required fields for the selected method aren't correctly set.


madleech:
Great stuff!

PR #49, opened 2016-03-23
Add saveJob() implementations

hghile:
This makes testing easier if nothing else.

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -1,0 +1,3 @@
+Thomas Oster
+
+Google Inc.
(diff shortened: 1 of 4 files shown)


AUTHORS
```diff
@@ -0,0 +1,3 @@
+Thomas Oster
+
+Google Inc.
```
madleech: Just curious where the Google comes from?

hghile: I work at Google, and our policies for contributing to open source software include adding this reference.

t-oster: So you are contributing while working? Or do you have to add this reference even for private work? I am just curious why Google is interested in VisiCut.

hghile: I'm contributing while working as part of my 20% non-primary-project time.
I'm interested in it because I don't want the proprietary windows-only laser driver software.

mgmax: looks like you forgot a few of the other authors. Could you please add some reference to the git log, like:
"And many others, see the git log for details." ?

PR #50, opened 2016-03-24
Replace use of LinkedList<Byte> with a wrapper around byte[]

hghile:
This makes handling of raster3d jobs much faster, and uses less memory allocation with less GC churn.
The amount of speedup depends on the size of the job.  For a 20x30cm 1000dpi job there was a 100x speedup.  For a full 40x60cm job it completed in 2 seconds, whereas before it still wasn't complete after half an hour.

### DIFF
--- a/src/com/t_oster/liblasercut/ByteArrayList.java
+++ b/src/com/t_oster/liblasercut/ByteArrayList.java
@@ -1,0 +1,153 @@
+/**
+ * This file is part of LibLaserCut.
+ * Copyright (C) 2016 Google Inc.
+ *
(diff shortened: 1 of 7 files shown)


t-oster:
Nice. Could you rebase the pull request into one commit?
Also is the copyright header mandatory? Right now there is a check in the build server which wants all files to have the same copyright header.

hghile:
I couldn't get git to cooperate with me, so I'll make a new pull request with a single commit and an updated header (I followed the model of additional copyright from GoldCutHPGL)
You can just drop this pull request.

PR #59, opened 2016-08-02
New dithering algorithms: Halftone and Brightened Halftone

mgmax:
These dithering algorithms are more robust against inaccuracies caused by real-life machines and materials.

- Halftone: a 8x8 ordered dithering with a dot-cluster pattern

  Because the pixels cluster together and form dots of varying size, it works better than other algorithms with finer pixel patterns.

- Brightened halftone: Halftone with an applied brightness correction for engraving on acrylic

  Similar to the "Brighten" dithering profile offered by Epilog, values chosen by trial-and-error to look good for engraving (inverted) photos on acrylic

### DIFF
--- a/test.sh
+++ b/test.sh
@@ -8,10 +8,12 @@ then
 	exit 1
 fi
 HEADERSIZE=$(wc -l < copyrightheader)
+# filter out the line that contains year, author name(s) and at least one email adress enclosed in <>
+IGNORE_AUTHOR_LINE_REGEXP='^ \* Copyright \(C\) 20[0-9]{2}.*<.*@.*>.*$'
 ERRORS=0
 for f in $(find src -name '*.java')
 do
-	if ! cat $f|head -n $HEADERSIZE|diff - copyrightheader > /dev/null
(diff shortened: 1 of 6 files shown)


t-oster:
Hi, is it a problem for you to change the license header like in https://github.com/t-oster/LibLaserCut/blob/develop/src/com/t_oster/liblasercut/drivers/MakeBlockXYPlotter.java? I have a `test.sh` script which checks for the presence of the header and it does not use any regular exprpessions, so it fails on you e-mail address.

test.sh
```diff
@@ -8,10 +8,14 @@ then
 	exit 1
 fi
 HEADERSIZE=$(wc -l < copyrightheader)
+# filter out the line that contains year, author name(s) and at least one email adress enclosed in <>
+IGNORE_AUTHOR_LINE_REGEXP='^ \* Copyright \(C\) 20[0-9]{2}.*<.*@.*>.*$'
+exit 1
```
t-oster: exit 1 ?

mgmax:
I fixed the obvious problem :-)
The remaining CI errors should now be due to #61.

t-oster:
I had something similar in VisiCut some time. I added some magic to .travis_yml, so please rebase in order to retry the build.

mgmax:
now rebased against #62 and happily working in travis.

PR #63, opened 2016-09-23
Add absolute speed option for Smoothieboard driver

pelrun:
Adds new Smoothie driver that changes the meaning of the Speed variable from percent to mm/min.

Needs more work, don't merge just yet.

https://github.com/t-oster/VisiCut/issues/381

### DIFF
--- a/src/com/t_oster/liblasercut/FloatPowerAbsSpeedFocusProperty.java
+++ b/src/com/t_oster/liblasercut/FloatPowerAbsSpeedFocusProperty.java
@@ -1,0 +1,89 @@
+/**
+ * This file is part of LibLaserCut.
+ * Copyright (C) 2011 - 2014 Thomas Oster <mail@thomas-oster.de>
+ *
+ * LibLaserCut is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU Lesser General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * LibLaserCut is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+ * GNU Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public License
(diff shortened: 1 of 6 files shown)


pelrun:
Thinking about it more, since this doesn't update the profiles (i.e. toggling it breaks everything), it might be better to have it as a separate driver rather than a driver option, to force the profiles to be rewritten.

mgmax:
As this is also applicable to all other GenericGcode based drivers, it would perhaps be better to make this a configuration option inside GenericGcodeDriver, named something like "Absolute speed (change will break existing settings)". It would also make the code easier to understand, if setSpeed just contains an if/else logic.

The configuration dialog even allows changing the driver, so we cannot prevent the user from switching the option anyway.

The Epilog driver is not based on Gcode, so it would still work as before.

pelrun:
I don't like having an option break all the profiles, especially when you can't then tell which ones have been updated. But I'm not happy with having multiple drivers either. The option needs to either auto-update the profiles, or just change how they're displayed. I'll look at doing the latter.

madleech:
Cool feature, I'd be very interested in having this as a general option for all gcode drivers.

t-oster:
I am not sure, if it is such a problem that it breaks the profiles, because if you are messing with the driver settings, you should know what you are doing anyway... 
....@madleech nice avatar....

pelrun:
It's still a UI failure. And fixing those has been the main focus of my work on VisiCut.

t-oster:
Hi, what is the state of this patch? I think adding it to the generic gcode driver is a good option

pelrun:
I'm not happy with it, and I didn't get far trying to resolve the issues that popped up. The big one was that the displayed text for the "power/speed/focus" options is magic and used as a key for serialisation. So I couldn't readily just change those dynamically to include units, because it breaks everything.

It did occur to me you can be super sneaky and just change the max speed in the driver to 6000 mm/min, at which point the percentage is also numerically identical to the speed in mm/s...

t-oster:
but it will only allow values between 0 and 100, right?

pelrun:
It's just a trick, not really a good solution :)

pelrun:
I'm having another stab at this, although the implementation is significantly different now - so I'll close this PR and raise a new one if/when I get it working.

PR #70, opened 2016-11-21
full spectrum driver

volvonk:
Adds support to full spectrum laser cutters, supports only cutting, uses a fixed speed interpolator, which is good enough for low speeds, there are some things to improve(see TODO list in FullSpectrumCutter.java) but at least the protocol and building of packets are already implemented.

### DIFF
--- a/src/com/t_oster/liblasercut/drivers/FullSpectrumCutterProperty.java
+++ b/src/com/t_oster/liblasercut/drivers/FullSpectrumCutterProperty.java
@@ -1,0 +1,201 @@
+/**
+ * This file is part of LibLaserCut.
+ * Copyright (C) 2011 - 2014 Thomas Oster <mail@thomas-oster.de>
+ *
+ * LibLaserCut is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU Lesser General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * LibLaserCut is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+ * GNU Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public License
+ * along with LibLaserCut. If not, see <http://www.gnu.org/licenses/>.
+ *
+ **/
+package com.t_oster.liblasercut.drivers;
+
+import com.t_oster.liblasercut.LaserProperty;
+
+/**
+ * Handles parameters for full spectrum jobs.
+ * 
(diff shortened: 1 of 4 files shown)


src/com/t_oster/liblasercut/drivers/FullSpectrumCutterProperty.java
```diff
@@ -0,0 +1,201 @@
...
+  {
+    if ("power".equals(name))
+    {
+      return (Float) 100f;
+    }
+    else if ("speed".equals(name))
+    {
+      return (Float) 100f;
+    }
+    else if ("focus".equals(name))
```
t-oster: this also?

src/com/t_oster/liblasercut/drivers/FullSpectrumCutterProperty.java
```diff
@@ -0,0 +1,201 @@
...
+  {
+    if ("power".equals(name))
+    {
+      return (Float) 0f;
+    }
+    else if ("speed".equals(name))
+    {
+      return (Float) 0f;
+    }
+    else if ("focus".equals(name))
```
t-oster: is this a leftover from copy-paste?

src/com/t_oster/liblasercut/drivers/FullSpectrumCutterProperty.java
```diff
@@ -0,0 +1,201 @@
...
+
+/**
+ * Handles parameters for full spectrum jobs.
+ * 
+ * Based upon PowerSpeedFocusProperty by Thomas Oster <thomas.oster@rwth-aachen.de>
+ * 
+ * @author Volkan Vonk <vol.vonk@yandex.com> 
+ */
+public class FullSpectrumCutterProperty implements LaserProperty
+{
```
t-oster: This class could be used on other devices and should thus be named `FloatPowerSpeedProperty` in order to be consistent with the other properties.

src/com/t_oster/liblasercut/drivers/FullSpectrumCutterProperty.java
volvonk: yeah, sorry about that

src/com/t_oster/liblasercut/drivers/FullSpectrumCutterProperty.java
volvonk: yeah, sorry again

volvonk:
Hello, thank you, i made the requested changes
The driver it is not complete, because it lacks of a acc/dec stage so for high speeds 60< the inertia begins to degrade the job quality, though very long and complex jobs may be performed with great accuracy at lower speeds. I think would be better to use lower speed settings by default maybe 600 or 500 steps/sec, what do you think about this?
Also the driver doesnt support engraving, but i may add some insights about engraving packets.

PR #82, opened 2018-03-16
Switch vectoroptimizers from LinkedList to ArrayList for memory and speed reasons.

Hueburtsonly:
* The optimizers and methods of Element have widespread use of .get(index) within loops, so switching from LinkedList (which is O(n) for arbitrary index lookup) to ArrayList (O(1)) will improves most of these methods from O(n^2) to O(n).

* Since the Point object is so small, the per-element memory overhead of LinkedList is larger than the extra capacity that an ArrayList might have.

### DIFF
--- a/src/com/t_oster/liblasercut/vectoroptimizers/VectorOptimizer.java
+++ b/src/com/t_oster/liblasercut/vectoroptimizers/VectorOptimizer.java
@@ -23,7 +23,8 @@ import com.t_oster.liblasercut.VectorCommand;
 import com.t_oster.liblasercut.VectorPart;
 import com.t_oster.liblasercut.platform.Point;
 import com.t_oster.liblasercut.platform.Rectangle;
-import java.util.LinkedList;
+import java.util.ArrayList;
+import java.util.Collections;
 import java.util.List;
 
 /**
@@ -47,28 +48,17 @@ public abstract class VectorOptimizer
 
     LaserProperty prop;
     Point start;
-    List<Point> moves = new LinkedList<Point>();
+    List<Point> moves = new ArrayList<Point>();
 
-  public boolean equals(Element e)
+    public boolean equals(Element e)
     {
-      if(this.moves.size()==e.moves.size())
+      if (!this.start.equals(e.start) )
       {
-        if (!this.start.equals(e.start) )
-        {
-          return false;//start point differs
-        }
-        
-        for (int j = 0; j < this.moves.size(); j++)
-        {
-          if (!this.moves.get(j).equals(e.moves.get(j)))
-          {
-            return false;//one move point differs
-          }
-        }
+        return false;//start point differs
       }
-      else
+      if(!this.moves.equals(e.moves))
       {
-        return false;
+        return false;//move lists are different
       }
       return true;
     }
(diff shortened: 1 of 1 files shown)


t-oster:
Hi, i dont think there is an element overhead since only references are stored imho. However are you sure the equals mehtod compares elements or just If they point to the same objects?

t-oster:
Well, does anybody know about how inefficient we are talking about? Any speed measurements with real-world jobs? Are we talking seconds? or minutes? @Hueburtsonly did you check your implementation agains the current version?

Hueburtsonly:
> Hi, i dont think there is an element overhead since only references are stored imho

You're correct that the size of the Point type is irrelevant, my bad, but my overarching point still stands: An element in a LinkedList stores _three_ references: the object, and two links (forwards and back). By comparison, each element of an ArrayList is indeed just a reference, with at most a 1.5x over-capacity of elements. So, roughly speaking, a LinkedList always consumes at least twice as much memory.

> The output of divide() is passed on to VectorOptimizer.sort(). The implementation NearestVectorOptimizer.sort() calls remove() frequently, which is very inefficient for ArrayList (O(n) for ArrayList vs O(1) for LinkedList).

No, LinkedList.remove(index) is O(n): O(n) to traverse to the given index, plus O(1) to do the actual deletion. [[ref]](https://stackoverflow.com/questions/33987542/time-complexity-of-deletion-in-a-linked-list)

Indeed, ArrayList.remove(index) is also O(n): "O(1)" to "traverse" to the given index, plus O(n) to shuffle the remaining elements down.

So while Iterator.remove() on a LinkedList is O(1), that's only actually O(1) if you have an iterator primed and ready to go. Since, AFAICT, it's impossible to clone an iterator, I think it's very tricky to perform a linear scan through all the points to find the nearest, while maintaining a ready-to-go iterator to perform the O(1) delete. Also, it seems rather irrelevant given that this came after a O(n) scan.

There are ways to get O(1) removes from an ArrayList too, if you're OK with discarding order (which you probably are if you're removing an item from the middle of a list). In short, always remove the last item (O(1)) and use it to replace item that you _actually_ want to delete (O(1)). But that's messy code too; I would consider that premature optimization unless a real performance problem was demonstrated.

I know the feeling, LinkedLists are so lovely and elegant, at least conceptually. The trouble is that you have to tread _very_ carefully, or else you'll drop in a get(index) or remove(index) and your performance is shot compared to even naively-written ArrayList impls. I'd suggest we go ahead and switch over to ArrayList now because that's what all the algorithms are optimized for at the moment, and in the (IMHO somewhat unlikely) event that a significant better algo that relies on LinkedList comes along, we can switch back? Much easier and safer to just change the type.

> The deduplication in DeleteDuplicatePathsVectorOptimizer could also be rewritten...

I agree, I was going to do something similar in a future change (was planning to just use HashSet; I hadn't heard of LinkedHashSet before, thanks for the hint!)

> Did you check your implementation agains the current version?

Speed measurements using a 40x40 grid of squares:

```
                    (ms) Before  After   Ratio
FILE                       0.00   0.00     --
NEAREST	               2,137.47  35.72   59.8x
INNER_FIRST               17.58  13.23    1.3x
SMALLEST_FIRST            14.45   7.70    1.9x
DELETE_DUPLICATE_PATHS 9,897.70  59.79  165.5x*

* Can be improved in other ways too as pointed out by mgmax.
```

t-oster:
I will update VisiCut in a few hours

Issue #83, opened 2018-03-18
Joining of contiguous elements


Hueburtsonly:
There's a suboptimality with the way that Visicut/LibLaserCut handles INNER_FIRST optimization of certain input files where nominally closed loops are broken up into individual line segments, which are then cut according to Y position, which wastes a lot of time jogging about. I have implemented a fix, a pre-processing step in InnerFirstVectorOptimizer, which joins up elements with matching endpoints (and props) into larger elements -- most importantly, if a bunch of elements make a simple loop, it will make a single element which is the whole loop. This has two benefits:

* Jog time greatly reduced

* Watching the laser cutter, it actually _looks_ like an inner-first algorithm (i.e., it will avoid a lot of confusion and annoyance I've seen from people saying Visicut is "not doing what I asked" even though it sort of is, if only on a technicality)

Before starting a pull request, I wanted to ask first how you would like this to be presented to the user. Although the change will increase CPU time, the increase will be much smaller than the decrease I've introduced in #82. However, the memory usage will be considerably higher (don't have any hard numbers for you yet, but some constant factor more than current memory usage.) So, some options:

A. Simply modify INNER_FIRST to add the pre-processing.
B. Rename INNER_FIRST to INNER_FIRST_WITHOUT_LOOP_JOINING or INNER_FIRST_LITE or something like that, and add a new INNER_FIRST optimizer with the preprocessing step.
C. Add a new INNER_FIRST_WITH_LOOP_JOINING (or similar) optimizer.

I personally don't like C because I think 95% of users would be happy to see the new behaviour, and giving it a long intimidating name seems unfortunate. Take DELETE_DUPLICATE_PATHS as an example, I had no idea until I dug into the code that it uses NEAREST behind the scenes; I had always used NEAREST because I erroneously thought I had to make a choice between delete duplicates and having optimized pathing. So I'm in favour of option A, or B if we need to give users a way out in the so-far-hypothetical case of encountering memory issues with the pre-processing.

I don't claim that my change is the final say in how INNER_FIRST should be implemented, it's purely a step in the right direction (IMHO). The future work mentioned in comments inside InnerFirstVectorOptimizer is quite orthogonal to my change, but I'd also be happy to see my change replaced by a holistic solution.

> Is vectoroptimizers the right place to do this pre-processing when this is really a bug with the Visicut DXF importer/the DXF format*?

I think so.

[...]

PR #84, opened 2018-03-18
Repair broken polylines in INNER_FIRST optimizer.

Hueburtsonly:
Partially addresses t-oster/VisiCut#203 and mentioned in #83.

### DIFF
--- a/src/com/t_oster/liblasercut/platform/Point.java
+++ b/src/com/t_oster/liblasercut/platform/Point.java
@@ -61,4 +61,13 @@ public class Point
     hash = 29 * hash + this.y;
     return hash;
   }
+
+  public int compareTo(Point o)
+  {
+    if (x < o.x) return -1;
+    if (x > o.x) return 1;
+    if (y < o.y) return -1;
+    if (y > o.y) return 1;
+    return 0;
+  }
 }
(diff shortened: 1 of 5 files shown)


Hueburtsonly:
I'm aware the test is broken, I will fix it (might be a few days though).

Performance was checked using a 20x20 grid of 31-gons (12.4k Line Segments). To my surprise, the speed of INNER_FIRST on this test was improved from 32 ms to 20 ms, even though the code adds in a new step. Apparently, the joining step happens quickly enough, and reduces the number of Elements for the 4 sort stages in InnerFirstVectorOptimizer enough to gain a net improvement in speed. For comparison, NEAREST takes ~1800ms on the same test data.

I know that there have been suggestions to use a certain epsilon in testing for point equality; however:
* My algorithm uses binary search to find points, which is incompatible with fuzzy equality (in more than one dimension). I wanted to keep the algorithm at O(N log N) to absolutely ensure that I didn't introduce a significant performance regression.
* Floating point errors are deterministic (I know there are interesting exceptions to this statement, but I'm not sure how relevant they are here), so two coordinates that start off equal will stay precisely equal after, e.g., scaling and rotation operations. More specifically, various tests I've done have not produced any failures due to floating point errors, despite the shapes having lots of (nominally) irrational point coordinates (31-gon).
* This algorithm is failsafe, i.e., if it fails to identify a merging opportunity, the laser cutting job will still be performed successfully, just suboptimally.

If we see t-oster/VisiCut#203 continuing to occur in the field, we can revisit to make further improvements.

t-oster:
Thank you for your work. Its nice to see a skilled Java developer at work.

PR #85, opened 2018-03-19
Add support for hiding focus setting and switching autofocus off

hghile:
Not all epilog lasers support focusing from software, and not all autofocus-enabled cutters want to use autofocus all the time. Add options to control these.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,8 +1,6 @@
 language: java
 jdk:
-  - oraclejdk7
   - openjdk7
-  - openjdk6
 
 
 script:
--- a/src/com/t_oster/liblasercut/LaserJob.java
+++ b/src/com/t_oster/liblasercut/LaserJob.java
@@ -39,6 +39,7 @@ public class LaserJob
   private double startX = 0;
   private double startY = 0;
   private List<JobPart> parts = new LinkedList<JobPart>();
+  private boolean autoFocusEnabled = true;
 
   public LaserJob(String title, String name, String user)
   {
@@ -140,5 +141,18 @@ public class LaserJob
       startX = 0;
       startY = 0;
     }
+  }
+
+  /**
+   * Sets whether autofocus is enabled for this job.
+   * No-op for lasers that do not support it.
+   */
+  public void setAutoFocusEnabled(boolean b) {
+    autoFocusEnabled = b;
+  }
+
+  /** Indicates whether autofocus is enabled for this job, assuming the cutter supports it. */
+  public boolean isAutoFocusEnabled() {
+    return autoFocusEnabled;
   }
 }
--- a/src/com/t_oster/liblasercut/PowerSpeedFocusFrequencyProperty.java
+++ b/src/com/t_oster/liblasercut/PowerSpeedFocusFrequencyProperty.java
@@ -28,16 +28,15 @@ public class PowerSpeedFocusFrequencyProperty extends PowerSpeedFocusProperty
 {
 
   private int frequency = 5000;
-  private boolean useFocus = true;
 
   public PowerSpeedFocusFrequencyProperty()
   {
   }
 
   /** Make a new PowerSpeedFocusFrequencyProperty that optionally utilizes the focus setting */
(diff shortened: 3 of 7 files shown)


mgmax:
Does the property SoftwareFocusEnabled really need to be in LaserJob? LaserJob is very generic, some machines might even not be lasercutters (plotters etc.).

hghile:
I'm confused by your comment because SoftwareFocusEnabled is not in LaserJob.

mgmax:
Sorry for the confusion, I meant to say  isAutoFocusEnabled(), which was newly added. From a quick look at the code, I don't really get why that needs to be there and not just in the device driver settings.

hghile:
Ah, OK.  The problem with having it in device driver settings is that it's on or off at the device level rather than the job level, which is inconvenient.  Our laser cutter supports autofocus, and sometimes we want to run a job with autofocus and sometimes we do not.  Currently in order to change this setting one needs to go through options->lasercutters->select->edit->change->save->save which is a lot of clicks.  And when it's changed the next person coming along has no visibility into whether autofocus is on or off (my next change will put an autofocus option into the visicut main UI).
I considered adding this option into LaserCutter, but it didn't seem cleaner.  And putting into LaserProperty doesn't fit with how the option works.  Putting it in LaserJob was the best fit I could find, and anything that doesn't support autofocus can just ignore it.

hghile:
ping

t-oster:
I won't be here the next week due to private reasons.

rbowmaker:
I'm not familiar with the code, but just wanted to add my agreement that the user shouldn't need to go into the machine settings to turn off autofocus for a job. In a maker space with a few admins and lots of users;

- Users can very reasonably want to turn off autofocus for their job (for instance, if the workpiece is very thin and flexible and would be damaged by the autofocus probe; or a non-flat work piece), but the next user that comes along should see the autofocus set back to enabled by default.
- Whereas, machine settings are permanent, persistent across Visicut restarts and new jobs, and basically shouldn't be touched by non-admins.

On this basis, I strongly agree with hghile that having the focus setting _only_ modifiable in the machine settings is not ideal. It should probably still be there, but also overridable on a per-job basis.

hghile:
ping

t-oster:
Well the machine settings should specify if the device is capable of using autofocus, while the job settings says if it should be used for that job. So it's fine with me.

mgmax:
Sounds good. If we later add more settings and decide to change this interface to something more generic (property + value, like the device driver settings), will it affect compatibility of the PLF format for saving laser jobs?

t-oster:
I thought also, but the PLF format shall not contain any device specific settings, because it should be portable between devices. By definition it only contains information about the source files, their positioning and their mapping to "named" laserprofiles (it does not include the specific device settings, just the name like "cut"). Since the autofocus settings is definitively device specific it has nothing to do with the PLF.

t-oster:
However I just tested: VisiCut deserializes saved settings with Xstream, which will load new booleans added to classes as "false". This means that every Epilog-Lasercutter will have the SoftwareFocusSupported property to false, which is bad.
We could try to work around this but not so easy https://stackoverflow.com/questions/16060804/set-default-value-for-fields-not-in-xml-in-xstream.

A more elegant way woud be to invert the meaning of the booleans in a way that false is the valid default value like "SoftwareFocusNotSupported" and "HideFocus" instead of use focus. That way all existing settings would keep the current behavior and can then be altered if needed.

hghile:
Thanks for checking that.  I made the change.
I also found that visicut deserializes the laserproperties with xstream, and since the hideFocus setting is part of the laserproperties it just takes the deserialized one rather than honoring what comes from the device settings.  But I think I have a not terribly hacky fix for that in visicut.

hghile:
ping

t-oster:
Well, I think I will need to fix the clone-methods in Epilog Zing and Helix class to copy this, but I will do this. I don't think this breaks anything, so I will merge it.

t-oster:
You did nothin wrong on the clone method, but there is already a `copyProperties` method which makes the clone call smaller and more active

PR #91, opened 2018-09-09
Driver for LTT iLaser 4000

mgmax:
looking forward to a review. Works as far as we tested the last week.


t-oster:
Hi, thanks for your work. Looks good to me. Not sure if we should include your python tool because of the different License though (i am not an expert in the whole license stuff, I just took the LGPL because I basically understood it as do whatever you want with the code).
What do you think?

PR #94, opened 2018-12-20
Bugfix: Copying materials for Laos laser ignored hideFocus preference

mgmax:
The clone() method in LaosCutterProperty did not copy hideFocus, hideFrequency etc.

(IMO these should never have been made properties of the laser settings, but rather be read from the currently selected lasercutter.)

### DIFF
--- a/src/com/t_oster/liblasercut/drivers/LaosCutterProperty.java
+++ b/src/com/t_oster/liblasercut/drivers/LaosCutterProperty.java
@@ -28,10 +28,10 @@ import java.util.LinkedList;
  */
 public class LaosCutterProperty extends FloatPowerSpeedFocusFrequencyProperty {
   
-  private boolean hidePurge = false;
-  private boolean hideVentilation = false;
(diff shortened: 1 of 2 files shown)


mgmax:
I missed something with the engrave settings, that's fixed (and force-pushed) now. Sorry for the inconvenience.

How I tested:
- Load FAU FabLab settings
- near the material dropdown, click "plus" and choose "copy settings from existing material" -> Acryl, choose new name "testtest".
- select this new material "testtest", 3mm
- add some image for engraving, "engrave all"
- open laser settings, make sure that "focus", "ventilation" (and others) are hidden (Before: not hidden, now: hidden)
- change to "cut all"
- open lasersettings,  make sure that "focus", "ventilation" (and others) are still hidden  (Before: not hidden, now: hidden)
- done.


Regarding the code at LaosEngraveProperty.clone(): Unfortunately, we can't call super.clone() because it tries to copy all properties (not just those from the superclass) and doesn't understand the engrave-related properties.

PR #95, opened 2018-12-21
Add bottom-up engraving support (Epilog, currently)

buxtronix:
Bottom-up engraving allows engraving from the bottom of the job, which (for Epilogs) avoids having the extraction fan draw debris over the engraved portion.

Epilog's documentation: http://support.epiloglaser.com/article/8205/29949/using-the-bottomup-engraving-option-in-the-laser-dashboard

There's a partner PR coming for Visicut itself.

### DIFF
--- a/src/com/t_oster/liblasercut/Raster3dPart.java
+++ b/src/com/t_oster/liblasercut/Raster3dPart.java
@@ -45,6 +45,14 @@ public class Raster3dPart extends RasterizableJobPart
   public int getBitsPerRasterPixel() {
     return 8;
   }
+  
+   public void setBottomUp(boolean bottomUp) {
+    this.cutBottomUp = bottomUp;
+  }
+
+  public boolean getBottomUp() {
+    return this.cutBottomUp;
+  }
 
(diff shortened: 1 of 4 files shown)


mgmax:
It is good that you address this for Epilog lasers, but unfortunately your solution collides with the current solution for LAOS lasers: Please have a look at how it's done for LAOS lasers (and some others which also reuse the laos-specific property classes), where it is part of the laser settings per material (just like speed and power). LAOS and Epilog should be usable in the same way, so we should decide which way is better and then make it the same for both drivers.

Personally, I would prefer a setting per material (which also matches what Epilog does in the Windows driver if you load preferences per material, and doesn't require changes in Visicut), but I'd be interested to hear other opinions.

t-oster:
Hi @buxtronix. Thanks for your work. However as @mgmax outlined if you make this setting part oft the laserproperty, it fits much better. Plus we dont need to touch visicuts code

buxtronix:
Canning this PR, finalising a replacement.

PR #96, opened 2019-01-01
Avoid errors with unsupported starting point

mgmax:
- Raise error message to avoid https://github.com/t-oster/VisiCut/issues/496
- Add dump utility and example files to help with troubleshooting the Epilog driver

### DIFF
--- a/src/com/t_oster/liblasercut/drivers/EpilogCutter.java
+++ b/src/com/t_oster/liblasercut/drivers/EpilogCutter.java
@@ -272,6 +272,17 @@ abstract class EpilogCutter extends LaserCutter
   @Override
   protected void checkJob(LaserJob job) throws IllegalJobException
   {
+    throw new AbstractMethodError("This should not be called.");
+  }
+
+  /**
+   * Check the laser job for obvious errors, such as physical dimensions
+   * @param job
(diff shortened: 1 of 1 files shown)


t-oster:
Shouldn't we investigate further before just blocking that Feature? I am pretty sure I cut some stuff with a centered starting point with an Epilog Zing

mgmax:
@t-oster is there any chance of getting this merged? The data we are sending doesn't match what the official Epilog driver does, so I don't think we are doing it right and we should rather disable the feature now.

If you don't want to merge, then I will send everything unrelated to the Epilog change as a new pull request.

t-oster:
If you can confirm that a vector job with starting point in the middle is not correctly executed when using the x/y off button to manually position the laser head, then we can merge, but I used it exactly that way on an epilog zing and it did work with negative coordinates. Unfortunately I do not live in Aachen anymore so I don't have access to the zing anymore

mgmax:
@t-oster I also don't have the possibility to test it right now. How about changing the error message into a warning like "This may not work. Please report if it works at https://github.com/t-oster/VisiCut/issues/496 " ?

t-oster:
Yes, I could live with that.

mgmax:
Okay, so now you should be able to merge this and #100.

PR #98, opened 2019-05-01
LaserScript: add prompt() function for user input

mgmax:
For now, this function is only a placeholder which is overridden by the VisiCut GUI. (pull request to be submitted soon)
If someone needs it, it could also be extended to support commandline (stdin) input.

Tested on Linux with the corresponding VisiCut pullrequest.

### DIFF
--- a/src/com/t_oster/liblasercut/laserscript/LaserScriptBootstrap.js
+++ b/src/com/t_oster/liblasercut/laserscript/LaserScriptBootstrap.js
@@ -41,3 +41,18 @@ function echo(text)
 {
   _instance.echo(text);
 }
+
+function prompt(title, defaultValue)
+{
+  return _instance.prompt(title, defaultValue);
+}
+
+function promptFloat(title, defaultValue)
+{
+  var result = parseFloat(_instance.prompt(title, defaultValue.toString()));
+  if (isNaN(result)) {
+    return defaultValue;
+  } else {
+    return result;
+  }
+}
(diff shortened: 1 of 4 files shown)


t-oster:
Nice one. However the promptFloat function seems arbitrary (why not a promptInt function etc) and could be done in the LS file itself. Also I think silently replacing wrong input with the default is confusing, but I don't want to nag, and it's always room for improvement.

mgmax:
Thanks for merging. There's certainly a lot that could be done -- what I did is just the minimum working prototype so that it can be used for test patterns. I documented the limitations in the Wiki: https://github.com/t-oster/VisiCut/wiki/LaserScript#prompting-for-user-input

Let's see how much the feature is used and what turns out to be most needed.

What are your thoughts on a "standard library" for LaserScript? Currently, there's a lot of copy-paste code in our examples, e.g. for writing seven-segment text. Would it be a good idea to include these as a class in the LaserScript bootstrap javascript file? Or as separate libraries, with some kind of include statement?

t-oster:
That's a good question. Depends on what people are using LaserScript for. The question is when it makes sense to turn things around and use LibLaserCut from another language if you want complex programmatic lasercutting. I added laser script mainly for power/speed test patterns and parametric stuff (but for that parametric SVG is better https://github.com/t-oster/VisiCut/wiki/Parametric-SVG). An include function should not be that difficult and would not hurt, so why not.

Issue #81, opened 2018-03-04
Support for stock K40 controller boards


dlarue:
Lots of people are using the inexpensive K40 laser cutter and one of the initial requirements was to "upgrade" the controller board to RAMPS or Smoothie, etc to gain access to other sofware besides the Corel software.  This also allowed VisiCut usage( via LabLaserCut ) but now there is open source software called K40 Whisperer( open source python code ) which allows direct control of the K40 laser cutter and while it is not VisiCut, it runs on anything Python runs on and has become popular.

Has there been any consideration to add direct K40 control in VisiCut/LIbLaserCut now that there's example code showing how it can be done?

http://www.scorchworks.com/K40whisperer/k40whisperer.html
https://github.com/dimsumlabs/k40_whisperer

morriswinkler:
@t-oster Have there been any updates regarding that question, did you talk back to the author of k40 whisperer?

t-oster:
Hi, here ist the answer: https://github.com/dimsumlabs/k40_whisperer/issues/1

tl;dr: It is possible, but no one has time to develop it.
However I guess it should not take more than 10-15 dev hours (just guessing). So if you get enough people to fund this (and probably a K40 for testing), I could implement this during my work time (120€/dev hour).

tatarize:
I'm a good way into writing a stock java driver for the Nano-K40 boards. 
https://github.com/K40Nano/Jk40

I spent a good while looking at the Whisperer code, and playing with the device itself itself and writing documentation for the whole thing: https://github.com/K40Nano/K40Nano/wiki and wrote a python library by extracting the important stuff from the tangle of whisperer code and giving python folks an api ( https://pypi.org/project/k40nano/ ).

I needed to rewrite the core elements to make it async and I wanted to allow more permissive licenses, which meant rewriting everything from scratch, so I switched to Java because I'm better at it, which seems rather serendipitous, given the language here.

So what exactly is needed to fully interface the stuff? I looked through the various drivers and I don't see any that are using usb connections.  At the core you need to connect through the USB to the read the status endpoint and writing LHYMICRO-GL code to the data endpoint. I spent the better part of month solving the speedcodes, and the code the laser uses. So that stuff is all pretty reasonable. And within my wheelhouse. Would putting that together be basically enough to do everything?

mgmax:
To be honest, which drugs did the manufacturer take when designing that data format? ;-)

We don't have any USB interface code in LibLaserCut, so you will need to bring your own. 

[...]

PR #101, opened 2019-06-15
Adds support for stock K40 board. M2-Nano and related.

tatarize:
This should be the vector part of the K40 driver.

I successfully ran a loaded project on my Stock Firmware K40 using Visicut on a Raspberry Pi.


tatarize:
With regards to https://github.com/t-oster/LibLaserCut/issues/81

src/com/t_oster/liblasercut/drivers/K40NanoDriver.java
```diff
@@ -0,0 +1,1484 @@
...
+        exit_compact_mode();
+        execute();
+      }
+      jobber.close();
+      jobber = null;
+    }
+
+    void setPower(double power)
+    {
+      this.power = power;
```
mgmax: it should be documented that setPower does nothing because the device doesn't support it

src/com/t_oster/liblasercut/drivers/K40NanoDriver.java
```diff
@@ -0,0 +1,1484 @@
...
+               * Change properties of current laser-actions (e.g. speed,
+               * frequency, power... whatever your driver supports)
+               */
+              LaserProperty prop = cmd.getProperty();
+              for (String key : prop.getPropertyKeys())
+              {
+                String value = prop.getProperty(key).toString();
+                if ("speed".equals(key))
+                {
+                  device.setSpeed(Double.valueOf(value));
```
mgmax: if I understand correctly, you're currently using the percentage value (speed 0...100) as mm/s, so there's a maximum of 100. That's okay -- choose as you like -- but should be documented in the code

Note that if you later change the meaning of "100", it will break existing saved laser settings, so you should rather decide now instead of in two years when everyone already has a list of saved material settings :-)

tatarize: Yeah, I was working to plug in pieces where they seemed to go and finish up the driver elements. I mostly skipped that part and used the default even though there's clearly other elements that need to go in there and bits that shouldn't be used.

tatarize:
Most of the separation of the driver interface and driver code is that I'm largely doing that work as a my personal project on the matter ( https://github.com/K40Nano/Jk40 ) and then pasting those classes back in there. My code is kinda needed to work independently for whatever other projects. And I wanted to match the general look of the drivers already there. So not clutter it up with a bunch of speculative classes. For 1400 lines it still looks kinda reasonable. 

More weird is the workflow of fixing stuff, compiling LibLaserCut copying the jar to a thumb drive, walking it over to the raspberry pi and running visicut there.

tatarize:
compare

I'm giving up on fixing that. I can't even find a character difference. In any event, Raster is in  but doesn't work and the properties are corrected.

mgmax:
Your file has windows line endings (CR LF) instead of unix line endings (LF), which are recommended with git. If you convert the files, e.g. using `dos2unix`, the error should be gone.

Why can't you run VisiCut on your development machine? That's what I usually do.

src/com/t_oster/liblasercut/drivers/K40NanoVectorProperty.java
```diff
@@ -0,0 +1,166 @@
...
+
+/**
+ *
+ * @author Tat
+ */
+public class K40NanoVectorProperty implements LaserProperty
+{
+  
+  static final String VAR_MM_PER_SECOND = "mm per second";
+  static final String VAR_D_RATIO = "diagonal speed adjustment";
```
mgmax: I'm not sure if this is a per-material property. If it only depends on the speed and not on the material itself, maybe it should rather be a machine property. Either as one constant, or two lists: a list of speeds and a corresponding list of diagonal factors.
For an example of how to use lists in the settings editor, see the LTT driver:
getProperty():
https://github.com/t-oster/LibLaserCut/blob/11bcce9051f8905103a0995923882a45a62071c6/src/com/t_oster/liblasercut/drivers/LaserToolsTechnicsCutter.java#L1823-L1830
setProperty():
https://github.com/t-oster/LibLaserCut/blob/11bcce9051f8905103a0995923882a45a62071c6/src/com/t_oster/liblasercut/drivers/LaserToolsTechnicsCutter.java#L1912-L1932

src/com/t_oster/liblasercut/drivers/K40NanoRasterProperty.java
```diff
@@ -0,0 +1,166 @@
...
+
+/**
+ *
+ * @author Tat
+ */
+public class K40NanoRasterProperty implements LaserProperty
+{
+  
+  static final String VAR_MM_PER_SECOND = "mm per second";
+  static final String VAR_RASTER_STEP = "raster step";
```
mgmax: I'm not sure about the purpose of "raster step", which seems to be a division factor to reduce the engrave DPI.
All other drivers just use the engrave DPI, which can be changed in VisiCut (slightly awkwardly via options - laser profiles - edit, but that's a different issue).

nbproject/project.properties
```diff
@@ -1,111 +1,127 @@
...
-# (you may also define separate properties like run-sys-prop.name=value instead of -Dname=value
-# or test-sys-prop.name=value to set system properties for unit tests):
-run.jvmargs=
-run.test.classpath=\
-    ${javac.test.classpath}:\
-    ${build.test.classes.dir}
-source.encoding=UTF-8
-src.dir=src
-test.src.dir=test
+annotation.processing.enabled=true
```
mgmax: You changed the file to Windows line endings. Please undo that.
See https://help.github.com/en/articles/dealing-with-line-endings for how to configure `git` to automatically convert line endings to LF before committing.

nbproject/private/private.properties
```diff
@@ -0,0 +1 @@
+user.properties.file=C:\\Users\\Tat\\AppData\\Roaming\\NetBeans\\8.2\\build.properties
```
mgmax: I don't think this file should be committed, rather added to .gitignore.

nbproject/project.properties
```diff
@@ -1,111 +1,127 @@
...
+dist.jar=${dist.dir}/LibLaserCut.jar
+dist.javadoc.dir=${dist.dir}/javadoc
+endorsed.classpath=
+excludes=
+file.reference.commons-net-3.1.jar=lib/commons-net-3.1.jar
+file.reference.corn-httpclient-1.0.12.jar=lib/corn-httpclient-1.0.12.jar
+file.reference.jna-4.0.0.jar=lib/jna-4.0.0.jar
+file.reference.js.jar=lib/js.jar
+file.reference.libusb4java-1.3.0-darwin-x86-64.jar=lib\\libusb4java-1.3.0-darwin-x86-64.jar
+file.reference.libusb4java-1.3.0-linux-aarch64.jar=lib\\libusb4java-1.3.0-linux-aarch64.jar
```
mgmax: you are using `\` as path separator. Not sure if this will break on Linux/MacOS, maybe better change it to `/` ?

src/com/t_oster/liblasercut/drivers/K40NanoRasterProperty.java
tatarize: It is, but it's oddly a division that occurs within the board. When in raster mode the speedcode has a G parameter it physically steps that number of mils when the right-left or top-bottom flag switches.  I added it because it was there, not sure how visicut does things which seems to be why the raster as a whole doesn't work. The K40 generally only has 1000 DPI, because it has stepper motors that step 1 mil. That property there makes the stepper motor step down by more than 1 row.  It might not be applicable.

src/com/t_oster/liblasercut/drivers/K40NanoVectorProperty.java
tatarize: Basically when both stepper motors trigger at the same time it takes a diagonal step which is longer than a purely ortogonal step. And the additional length is `sqrt(2) - 1` this means that the speed of the machine  ultimately depends, in part, on the angle it's drawing a line at. There's some other functional distances to expected idealized differences because since it takes tiny steps the the line drawing algorithm at the lowest level will result in some error giving your overall speed between 100% and 141% of the tickrate of the stepper motors. To try and minimize error there's a factor of addition ticks to wait between stepper motor pulses that is appended to the M2 and B2 boards. This tends to be 26% was their "auto" speed in the chinese software. But, this factor can be set to whatever, though really it shouldn't ever actually be changed. But, again, it's an actual factor the machine actually uses so I offered it up.

 If you are only cutting purely diagonal and purely orthogonal direction lines then it's better to use 0.41 there, as it will make the cut depths identical. 

chinese-sqrt2

tatarize: It's built into the machine. And does matter depending on the type of cut you are making. You could possibly want one value for one type of cut and a different one for another. I can't see it being popular, or even common, or anything but cryptic and ignored but the speedcode and diagonal adjustments are two different elements. I'm strongly lead to believe that this is where that property should end up. If it were machine based it you couldn't change it from class of cut to another. When there could be cause to cut things at 55mm/s with a diagonal speed addition of 0.26 because you're cutting a bezier curve and cutting at 55mm/s with a diagonal speed addition of 0.41 because all your cuts are purely orthogonal and diagonal. 

It's cryptic as heck even in the chinese software that gives you a random ass percent and no explanation of what it was, but since it modified the speed code it was analyzed, reverse engineered, and reproven. My research on the subject even made it into the K40 Whisperer since he hadn't reverse engineered the speed codes very well. And research might be needed to find values for circles etc, as these might actually have points of minimum error at different values.

tatarize:
I could run visicut on the development machine but it would involve either hauling an ATX case heavy computer or a laser cutter with a bucket of water and various pumps. Doesn't seem worth it for a project that I'm almost done with. 

* get raster working properly.
* implement a few other elements
* maybe add in a bell or whistle more. Especially if it requires deep knowledge of how the board works.
* send out a clarion call for K40 users to test this stuff.
* correct those issues if they deal with deep knowledge of how the board works.
* There might be some stuff to internally handle the usb errors, and issue warnings about the connection.
* Then it's all meh. Most people can do whatever coding, it's mostly the technical knowledge that is excessively rare.

A good project with a solid completion is a beautiful thing.

t-oster:
Thank you both. I am only briefly following, so @tatarize do you consider this driver ready for merging? @mgmax any objections?

mgmax:
I will have a look tomorrow. I hope that's okay.

tatarize:
It needs to pass a couple more benchmarks.

It needs to explicitly do raster encoding correctly. Currently its bugged and doing it right seems to require hamfisting a routine to see whether I  am further right or left than the last black pixel on the next line (but I can't directly tell from the helper code if I'm going right or left). The M2 board triggers a step when you flip right-left or left-right. So I can't step down to the next row unless I'm far enough to the extreme. But, I can't know which extreme.  And I had to read through the sources to make sure I was doing the stuff correctly and find a way around this defect. -- Hopefully done tonight.

And I need to try it on some larger project. Something that stresses it out and see whether the safeguards I put in place are at least robust enough to pass something more strenuous than a quick couple inches of vector work. I might need to catch some of the LibUsbExceptions in the driver itself and try to recover them. Unplugging the usb will, currently, kill the project even though technically it will lose absolutely no data. And I could just call execute() again if the problem was resolved and it would keep processing the buffer and queue.

tatarize:
Oh, and there's a strong preference that some other than me tries the code on their K40, and reports that it seems to work. If it fails, I'd hate to have it fail for many people in production rather than one person in testing. I might be doing something special to get it to work that I don't know about.

tatarize:
Raster works. But, it's fickle.  I think I'm messing up the driver stuff with regard to DPI and with regard to setting the speed. I naively assumed I would always be gifted the speed; I am not. I also figured I'd get a scaled image and not have to scale the DPI myself. I'm thinking I might have to set that to 1000 dpi only and hope that fixes things. Also, I apparently need to correct the movement of the raster element so it lines up with the vector stuff.

tatarize:
Tries to do this. Not sure why. My test version doesn't throw this error.

"Tried to load a laser-property of class class com.t_oster.liblasercut.PowerSpeedFocusFrequencyProperty, but lasercutter expects com.t_oster.liblasercut.drivers.K40NanoVectorProperty@5995e727
Trying to copy most values"

Then fails to run the raster, though it has rastered stuff before. Just need it to run to see that I have things lined up correctly when I set it to 1000dpi or nothing.

tatarize:
Raster works and resets positions correctly and runs fine. It doesn't however, work with images. It throws the above error of wanting PowerSpeedFocusFrequency. I mean I guess I could add some fake settings but I'm not very skilled with what that stuff does. The machine only has speed. I'm thinking the problem is expecting it to have those things, when it doesn't.

I never figured out the scaling with regard to the DPI. I could certainly step 2 lines and move by 2 spots and call that 500dpi for images. But, the machine only has 1000 dpi. And I'd need to understand that stuff better. I just said I had only 1000 DPI and it used that instead.

Also, the device tracks the positioning because the head is where it is, and it seemed like if you ran it a second time, you'd want that, but I'm thinking I should just rehome the device at the end of the job. Which seems reasonable enough and minor enough. Without doing that if the device rehomes it will be at 0,0 and think it's at 200 200 or whatever and then slam into the wall if you try going to 0,0.

I'll make it rehome after the job, and take another stab at DPI. But, the bulk of the stuff is done.

tatarize:
Fixed the DPI, rehomes after the job. Last cleanup done. Extended the timeouts with the USB. It can move for longer than 2 seconds. But, likely not 5 and the hello packet times out during a rapid move. Removed the raster_step element, tied it into DPI properly.

---

I'm officially out of things that need fixing. @t-oster I'm fine with it being pulled.  It seems to work. Emphasis on seems. Needs to be tried by other people and feedback gotten.

t-oster:
Thank you. I rebased all commits into one and merged it. I will release an updated VisiCut version now.

PR #107, opened 2019-06-22
Optimizes BlackWhiteRaster.

tatarize:


tatarize:
Clean version of #102. WhiteBlackRaster overtly corrects for incongruity with the meaning of "black". Delegates to the cleaner version of the class as part of piecemeal approach.

src/com/t_oster/liblasercut/BlackWhiteRaster.java
```diff
@@ -89,77 +88,128 @@ public BlackWhiteRaster(GreyscaleRaster src, DitherAlgorithm dither_algorithm) t
...
+  public BlackWhiteRaster(int width, int height, int bitDepth, int samplesPerPixel)
   {
-    int bx = x / 8;
-    int ix = 7 - (x % 8);
-    return ((raster[bx][y] & 0xFF) & (int) Math.pow(2,ix)) != 0;
+    this.raster = new RasterElement(width, height, bitDepth, samplesPerPixel);
   }
 
-  public void setBlack(int x, int y, boolean black)
+  public int getPixel(int x, int y)
```
t-oster: Could you add some method descriptions what exactly getPixel and setPixel do?

tatarize: Do note setBlack just got moved, it's down on line 168
`public void setBlack(int x, int y, boolean black)`

They delegate to the RasterElement which literally just sets the pixel. I mean I guess it sets the rightmost bitDepth bits of the integer into the given to the pixel space regardless of byte boundaries or other limitations. I could certainly do a better job documenting that.

But, I think including the delegate there was in error. The original BlackWhiteRaster didn't have a set pixel so this one might be better off without it. Though clearly, nothing is currently calling that code.

With the typical bitDepth of 1 this will simply set the pixel to 0 or 1 equal to (pixel & 1). also, getPixel (x,y) will return 0 or 1 depending on what that pixel's state is. It's basically the same as isBlack(x,y) at that point except that it doesn't decree that 1 is a black and 0 is white. It just gives the pixel's actual state. The function it's delegating to does this same thing for all bitDepths  (n <=32). Here the raster should be a 1 bit raster and thus just do what's expected for bitDepth 1. Namely set a 1 or 0 to the pixel in question or get a 1 or 0 from the pixel requested. I'll just add some documentation there. Shouldn't need much, it's called setPixel() and getPixel() and it sets and gets a pixel.

tatarize:
Trying to use the latest code for the driver without changing things I realized the BlackWhiteRaster needed to be able to return the RasterElement underlying it. The RasterBuilder function is great but requires that class type. And I documented the functions you asked about.

Rebasing this only needs to drop BlackWhiteRaster in the branch to replace BlackWhiteRaster. Since the other class already went through.

t-oster:
Do we need to test this with all existing drivers or should it be transparent to the outside?

tatarize:
The BlackWhiteRaster had entirely private internal structures. All the functions in here return the same data they returned in the old version. I simply delegated to a different class that does the same things in a more robust manner, thus being correct to replace the GreyRaster and AdapterBufferImage classes. So it should be a modular change. 

In addition, BlackWhiteRaster isn't actually directly called by any of the drivers.  Those use the helper functions in the RasterPart code. These elements of reading and writing in the object are covered in the test functions for BlackWhiteRaster and RasterPart.

The intent was to replace BlackWhiteRaster with RasterElement but that was too severe, so this merely delegates to the more robust code. Which will speed things a bit, but should categorically not break anything.

PR #109, opened 2019-06-22
Adds in RasterBuilder.

tatarize:
* Depends on RasterElement and AbstractLaserProperty. 
* Requires slightly reduced declarations on VectorCommand. 

RasterBuilder works as an iterable/iterator (you can use it in a for-loop) and is based on easily debuggable states as you would find in a state machine. It requires and allocates nearly no memory. It reuses the `VectorCommand` interface so already existing Vector code can be easily reused. Permits either axis rastering, starting in any of the four corners. 

Should work as a more robust replacement for the `LaserCutter.convertRasterizableToVectorPart()` function and related code, or any of the LaserCutter drivers implementation. Recycles the LaserProperty instance of CmdType.SETPROPERTY which it derives from the calling function via the `PropertiesUpdate`. 

The user defines what pixels should mean to them.

* Natively implements ProgressListener. 
* Natively includes overscan.
* Skips blank pixels based on the user's definition of blank. Making the scheme acceptable for both 3d rasters and black and white rasters.
* State-based implementation highly subject to debugging, and improvements.
* Agnostic with regard to the settings and LaserProperty the driver requires.
* Negligibly memory footprint.

```
RasterBuilder rb = new RasterBuilder(rp.image, new RasterBuilder.PropertiesUpdate() {
          @Override
          public void update(AbstractLaserProperty properties, int pixel) {
                properties.addProperty("color", pixel);
          }
        }, RasterBuilder.SERPENTINE_TRANVERSE_X_FROM_TOP_LEFT_SKIPPING_BLANK_LINES, 255, overscan);

rb.setProgressListener(pl);
rb.setOffsetPosition(rp.getMinX(), rp.getMinY());
        
for (VectorCommand cmd : rb) {
          //process the commands.
         // The exact value of pixel will be in the property "color".
}
```

### DIFF
--- a/src/com/t_oster/liblasercut/AbstractLaserProperty.java
+++ b/src/com/t_oster/liblasercut/AbstractLaserProperty.java
@@ -1,0 +1,259 @@
+/**
+ * This file is part of LibLaserCut.
+ * Copyright (C) 2011 - 2014 Thomas Oster <mail@thomas-oster.de>
+ *
+ * LibLaserCut is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU Lesser General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * LibLaserCut is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+ * GNU Lesser General Public License for more details.
+ *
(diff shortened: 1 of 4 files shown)


tatarize:
Note, it depends on RasterElement not the alterations to BlackWhiteRaster that rebases that on the improved class.  Even opting against touching BlackWhiteRaster this code could be utilized. It's setup to piecemeal replace the functionality with elegant debuggable code. Not to drop in replace anything. It's intended as a beachhead towards code cleanup, rather than asking for any even potentially breaking alterations.

PR #110, opened 2019-06-25
More Robust K40 Usb

tatarize:
Mostly adds in a routine to Wait for USB. On a failure to send the HELLO (get_status) packet it will shift into a holding pattern. If the device isn't found it will try to reconnect the device. If it fails for 30 seconds, it'll throw an error and abort the project. Updating the status to the progress listener.

This allows the user to turn the K40 off and have the project eventually die. And allows them to unplug and replug the USB plug a bunch if they want, it won't affect things. But, it will also save itself with regard to some more minor errors like mere timeouts.

During testing it hit several instances where the device stuttered and recovered fine.

### DIFF
--- a/src/com/t_oster/liblasercut/drivers/K40NanoDriver.java
+++ b/src/com/t_oster/liblasercut/drivers/K40NanoDriver.java
@@ -66,6 +66,8 @@ public class K40NanoDriver extends LaserCutter
   String board = "M2";
   boolean mock = false;
   PrintStream saveJob = null;
+  List<String> warnings = null;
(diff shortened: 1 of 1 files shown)


t-oster:
Does this depend on other PRs? I am sorry for the delays, I am not very active with regards to LaserCutting right now, so I am a little conservative regarding changes to existing drivers. This is your driver, so merging is not a problem.

tatarize:
This improvement doesn't depend on anything else and is isolated to just the K40Driver.

PR #111, opened 2019-06-27
Fix RasterBuild Overstep.

tatarize:
If the position of END is black, it will offer white pixels between END and the last actual pixel. It should rather always cut that off in order to truncate the blank extremes consistently.

### DIFF
--- a/src/com/t_oster/liblasercut/RasterBuilder.java
+++ b/src/com/t_oster/liblasercut/RasterBuilder.java
@@ -490,6 +490,7 @@ public class RasterBuilder implements Iterable<VectorCommand>, Iterator<VectorCo
   {
     getUpdatedPixelAtLocation();
     x_position = nextColorChangeHeadingLeft(x_position, y_position, end);
+    x_position = Math.max(x_position,end);
     commitPosition();
 
     if (end < x_position)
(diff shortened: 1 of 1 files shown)


pelrun:
The overscan isn't an error, it's there deliberately to prevent the last pixels in a line being cut off because the power dropped below the lasing threshold while the head was decelerating. If you don't want it/don't need it, you can always configure the overscan to 0 in the laser cutter settings...

tatarize:
Not the overscan, that was clearly a feature and is included in the RasterBuilder. The problem is that there is a bug wherein it's possible and happens about half the time to not truncate the scan. It literally oversteps to the end of the rasterline rather than stopping at the last writable pixel (plus the overscan). It was supposed to always stop there, but if there wasn't a transition point to to check, it would go to the next relevant point which was the rasterline end. Which was a marked error. About half the time the RasterBuilder code would incorrectly go to the end of the rasterline.

This is to fix the overstep error, not the overscan feature. The RasterBuilder makes implementing overscan for a particular driver easy. And should make it all work out. But, the feature to skip blank pixels was broken and overstepped.

It's not harmful and correct the code error. @t-oster should likely just approve it.

PR #112, opened 2019-06-27
Several Tweaks, K40 NanoDriver running with RasterBuilder.

tatarize:
Include #107 (still pending) and #111 (minor fix). And includes the changes from #110. But, mostly modifies the K40Driver to use raster builder and thus fundamentally skips blank pixels to the extremes.  And requires a change to RasterPart to expose the image to the drivers.


t-oster:
Please rebase

tatarize:
Yeah, with the other elements approved this one's dependencies on them seem like *way* more changes than they actually are, it basically just means slight changes to just the K40NanoDriver.

PR #115, opened 2019-07-20
Adds in RasterElement.Provider interface

tatarize:
This should be harmless as it only adds in an implementing interface and an unused scaffolding class.


tatarize:
This lays the foundations for using GreyscaleRaster implementations that also always implement `RasterElement.Provider` this should permit movement towards replacing `RasterizableJobPart.image` with a RasterElement itself. And removing any use of BufferedImage from `LibLaserCut`, currently only in `BufferedImageAdapter` which is currently tightly coupled and critical.

If a subgoal is to permit Android use of LibLaserCut, then that would mean lowering the importance of BufferedImageAdapter. Which would require making it possible to do the correct image manipulations in Visicut or a different library.

PR #116, opened 2019-07-29
Delegate to more Basal Classes. Generalize RasterPart.image

tatarize:
Since `BlackWhiteRaster` now implements `RasterElement.Provider` `BlackWhiteRaster` is no longer directly needed in the K40Driver. We can assume whatever the `RasterPart.image` is, it implements `RasterElement.Provider`.

RasterPart accepted only a `BlackWhiteRaster` than assigned it to GreyscaleRaster image. Accepting the baseclass permits a switch over.

However, in addition to this, internally `RasterPart.isBlack()` and `RasterPart.getRasterLine()` cast the `RasterPart.image` to `BlackWhiteRaster`. This delegate to `BlackWhiteRaster`, which now delegates those functions to `RasterElement`. The solution is directly delegate RasterElement. The delegation of `isBlack()` maintains its idiosyncrasy that 1 bit black is 0b1. The delegate to `getRasterLine()` gets the relevant line without regard to the BitDepth(). In RasterPart the bitdepth should always be `1` so this will work as expected.

Some test coverage for the getRasterLine() was added.

---

This should make it so that datastructure-wise, there is no functional difference between a BlackWhiteRaster and a 1bit GreyRaster.

Since the functions already delegated to BlackWhiteRaster, which delegated to RasterElement, the direct delegations with merged functionality should be necessarily correct. However, this does touch on some critical elements and should be verified, as it is not prima facie harmless.

### DIFF
--- a/src/com/t_oster/liblasercut/RasterPart.java
+++ b/src/com/t_oster/liblasercut/RasterPart.java
@@ -31,7 +31,7 @@ public class RasterPart extends RasterizableJobPart
   LaserProperty blackPixelProperty = null;
   LaserProperty whitePixelProperty = null;
 
-  public RasterPart(BlackWhiteRaster image, LaserProperty laserProperty, Point offset, double resolution)
+  public RasterPart(GreyscaleRaster image, LaserProperty laserProperty, Point offset, double resolution)
   {
     this.image = image;
     this.start = offset;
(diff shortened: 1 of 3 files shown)


t-oster:
Well, I created the same raster file with both versions and there was no difference, so I assume it works.

I will merge this for now, however I find that we may need much deeper cleanup. The reason I differenced between a BlackWhiteRaster and a GreyscaleRaster in the first place was because the Epilog driver has two different modes of operation for 1 and 8 bit raster depth.
We should probably merge all the classes to an allmighty Raster with a bit depth and add a driver property `supportedRasterBitDepths`. But this would be a huge refactoring, so maybe we should first analyze which devices do support which bit depth anyway.

PR #122, opened 2019-11-07
remove unneded newline and unneded the

matkoniecz:
I am 90% sure that "the" was not needed
I am 99% sure that the newline was not needed


matkoniecz:
note to self: after processing this PR open a new one for https://github.com/t-oster/LibLaserCut/issues/123

PR #128, opened 2019-12-24
GenericGcodeDriver: Add setting POST_HTTP_UPLOAD_GCODE to send GCode after HTTP upload

pimsierhuis:
At Fablab Amersfoort, we want to implement a workflow in which a job isn't automatically executed after uploading it to our Smoothieboard-controlled lasercutter. Instead, we want to implement a physical button to start cutting after uploading the job. Because the filename uploaded by Visicut is different each time and Smoothieboard isn't remembering what file is uploaded last, this isn't possible with the current version of Visicut and Smoothieboard (as far as we know). Because Smoothieboard allows "selecting" a file wth GCode "M23 <filename>" and after that executing it with "M24", we need Visicut to execute "M23 <filename>" after uploading the file to Smoothieboard. Thinking about this, we found that probably the most acceptable way for the Visicut developers would be a generic setting for executing the given GCode after the HTTP-upload. This is what this PR is about.

### DIFF
--- a/src/main/java/de/thomas_oster/liblasercut/drivers/GenericGcodeDriver.java
+++ b/src/main/java/de/thomas_oster/liblasercut/drivers/GenericGcodeDriver.java
@@ -55,6 +55,7 @@ public class GenericGcodeDriver extends LaserCutter {
   protected static final String SETTING_FLIP_Y = "Flip Y Axis";
   protected static final String SETTING_HTTP_UPLOAD_URL = "HTTP Upload URL";
   protected static final String SETTING_AUTOPLAY = "Start Job after HTTP Upload";
+  protected static final String SETTING_POST_HTTP_UPLOAD_GCODE = "GCode to send after HTTP Upload";
   protected static final String SETTING_LINEEND = "Lineend (CR,LF,CRLF)";
   protected static final String SETTING_MAX_SPEED = "Max speed (in mm/min)";
   protected static final String SETTING_TRAVEL_SPEED = "Travel (non laser moves) speed (in mm/min)";
@@ -161,6 +162,12 @@ public class GenericGcodeDriver extends LaserCutter {
   {
     this.autoPlay = autoPlay;
   }
+
+  private String postHttpUploadGcode = "";
+
+  public String getPostHttpUploadGcode() { return postHttpUploadGcode; }
+
+  public void setPostHttpUploadGcode(String postHttpUploadGcode) { this.postHttpUploadGcode = postHttpUploadGcode; }
 
   protected String supportedResolutions = "100,500,1000";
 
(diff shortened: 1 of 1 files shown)


src/main/java/de/thomas_oster/liblasercut/drivers/GenericGcodeDriver.java
```diff
@@ -549,8 +556,23 @@ protected void http_upload(URI url, String data, String filename) throws IOExcep
...
   {
+    String command = "play /sd/"+filename;
+    http_command(command);
+  }
+
+  protected void http_commands(String commands, String filename) throws IOException, URISyntaxException
+  {
+    for (String command : commands.split(","))
+    {
+      command = command.replace("$filename", filename);
```
matthijskooijman: Looking at this, I wondered whether this should not be `$jobname`, e.g. like used [here](https://github.com/t-oster/VisiCut/blob/b498e3a129e1cff4aa3657bd93751b10a52d1d13/src/main/java/de/thomas_oster/visicut/gui/MainView.java#L2321). However, it seems that these are actually slightly different values, since the liblasercut adds the `.gcode` extension, transforming the the jobname into a filename [here](https://github.com/t-oster/LibLaserCut/blob/d6728e8df802d631fd9f1e54d73fa45a2a26d04b/src/main/java/de/thomas_oster/liblasercut/drivers/GenericGcodeDriver.java#L823).

IOW, using `$filename` here seems good to me.

src/main/java/de/thomas_oster/liblasercut/drivers/GenericGcodeDriver.java
```diff
@@ -549,8 +556,23 @@ protected void http_upload(URI url, String data, String filename) throws IOExcep
 
   protected void http_play(String filename) throws IOException, URISyntaxException
   {
+    String command = "play /sd/"+filename;
+    http_command(command);
+  }
+
+  protected void http_commands(String commands, String filename) throws IOException, URISyntaxException
+  {
+    for (String command : commands.split(","))
```
matthijskooijman: Is , a good separator here? That prevents using a , inside the gcode commands sent? Ideally, you would just use a newline, but I think the GUI does not allow that?

src/main/java/de/thomas_oster/liblasercut/drivers/GenericGcodeDriver.java
pimsierhuis: That's exactly my reasoning why i finally came up with $filename :)

src/main/java/de/thomas_oster/liblasercut/drivers/GenericGcodeDriver.java
pimsierhuis: The reason for a comma as a separator is that this is also used with preJobGcode and postJobGcode

matthijskooijman: Good point, better to use comma here for consistency as well, then.

t-oster:
@pimsierhuis does this patch change the default behavior to not execute the job after uploading?

matthijskooijman:
@t-oster, no, the existing "Execute job after upload" boolean preference is still applied as before (with the default unchanged), separately from this new preference. When both are configured, both will be executed. If you, like us, need select after upload rather than execute after upload, you need to uncheck the execute preference *and* add an M23 select command.

However, looking at the code again I see that in tat case the "play" command is executed first, and then the custom gcode. I'm not sure in which cases you would want to combine these exactly, but intuitively I would think that reversing their order might be better? So you first execute custom gcode after the upload (maybe for additional setup before starting playback) and then you start play. I guess if you need to do stuff *after* play starts, you can use the other existing preference that puts gcodes inside the uploaded file anyway. @pimsierhuis, what would you think about this?

t-oster:
I agree that order makes more sense. Please update then I will release a new version

pimsierhuis:
I completely agree with you, that was actually my intention but somehow it worked out this way.. I updated the branch. A release would be very welcome. Thanks for the quick action!

t-oster:
I guess I need a rebased branch and a new pull request

pimsierhuis:
Done

PR #136, opened 2020-01-29
PPI power modulation.

tatarize:
Replace the linedraw algorithm using Zingl-Bresenham rather than Bresenham.
Appended carry-forward power modulation.
Corrected a LibUsb Error code.
Fixed spelling "packst" -> "packet"
Modified driver to provide 'power' (0, 1000) as a laser property.

### DIFF
--- a/src/main/java/de/thomas_oster/liblasercut/drivers/K40NanoDriver.java
+++ b/src/main/java/de/thomas_oster/liblasercut/drivers/K40NanoDriver.java
@@ -16,10 +16,10 @@
  * along with LibLaserCut. If not, see <http://www.gnu.org/licenses/>.
  *
  **/
+
 package de.thomas_oster.liblasercut.drivers;
 
 import de.thomas_oster.liblasercut.AbstractLaserProperty;
-import de.thomas_oster.liblasercut.BlackWhiteRaster;
 import de.thomas_oster.liblasercut.IllegalJobException;
 import de.thomas_oster.liblasercut.JobPart;
 import de.thomas_oster.liblasercut.LaserCutter;
@@ -54,8 +54,9 @@ import org.usb4java.LibUsbException;
 
 public class K40NanoDriver extends LaserCutter
 {
+
   private static final String VAR_MM_PER_SECOND = "mm per second";
-  private static final String VAR_D_RATIO = "mm per second";
+  private static final String VAR_D_RATIO = "diagonal ratio";
   private static final String SETTING_BEDWIDTH = "Laserbed Width";
   private static final String SETTING_BEDHEIGHT = "Laserbed Height";
   private static final String SETTING_BOARD = "M2, M1, M, B2, B1, B, A, board selection";
@@ -77,8 +78,7 @@ public class K40NanoDriver extends LaserCutter
 
   /**
    * This is the core method of the driver. It is called, whenever LibLaseCut
-   * wants
-   * your driver to send a job to the lasercutter.
+   * wants your driver to send a job to the lasercutter.
    *
    * @param job This is an LaserJob object, containing all information on the
    * job, which is to be sent
@@ -147,7 +147,7 @@ public class K40NanoDriver extends LaserCutter
         device.move_absolute(sx, sy);
         int step_size = (int) (1000.0 / p.getDPI());
         device.setRaster_step(step_size);
-        RasterElement element = ((RasterElement.Provider)rp.getImage()).getRaster();
+        RasterElement element = ((RasterElement.Provider) rp.getImage()).getRaster();
         RasterBuilder rasterbuild = new RasterBuilder(element, new RasterBuilder.PropertiesUpdate()
         {
           @Override
@@ -157,9 +157,9 @@ public class K40NanoDriver extends LaserCutter
           }
         }, 0, 0, 0);
         rasterbuild.setOffsetPosition(rp.getMinX(), rp.getMinY());
-        
+
         int pixel = 0;
-        device.raster_start();  
+        device.raster_start();
         for (VectorCommand cmd : rasterbuild)
         {
           if ((cmd.getType() == VectorCommand.CmdType.MOVETO) || ((cmd.getType() == VectorCommand.CmdType.LINETO) && (pixel == 0))) //treat moveto with pixel 0 as a lineto.
@@ -168,8 +168,9 @@ public class K40NanoDriver extends LaserCutter
             int y = (int) (cmd.getY() * (1000 / p.getDPI()));
             int dx = x - device.x;
             int dy = y - device.y;
-            if (dy > device.raster_step) {
-              device.move_absolute(x, y-device.raster_step);
+            if (dy > device.raster_step)
+            {
+              device.move_absolute(x, y - device.raster_step);
               //if we're moving in the y direction, but more than the raster step,
               //we still need to h_switch to change the directionality. But, that will
               //step, so we go down to where the raster-step will put us on the correct line.
@@ -179,7 +180,7 @@ public class K40NanoDriver extends LaserCutter
               device.h_switch();
               device.y += device.raster_step;
             }
-            
+
             device.move_absolute(x, y);
             device.execute();
           }
@@ -213,7 +214,7 @@ public class K40NanoDriver extends LaserCutter
         int total = vp.getCommandList().length;
         for (VectorCommand cmd : vp.getCommandList())
         {
-          this.progress.taskChanged(this, "Vector Part");  
+          this.progress.taskChanged(this, "Vector Part");
           this.progress.progressChanged(this, (100 * i++) / total);
           switch (cmd.getType())
           {
@@ -290,6 +291,7 @@ public class K40NanoDriver extends LaserCutter
   {
     AbstractLaserProperty property = new AbstractLaserProperty();
     property.addPropertyRanged("mm per second", 30f, 0.4f, 240f);
+    property.addPropertyRanged("power", 1000, 0, 1000);
     property.addPropertyRanged("diagonal speed adjustment", 0.2612f, 0f, 1f);
     return property;
   }
@@ -299,9 +301,9 @@ public class K40NanoDriver extends LaserCutter
   {
     AbstractLaserProperty property = new AbstractLaserProperty();
     property.addPropertyRanged("mm per second", 60f, 5f, 500f);
+    property.addPropertyRanged("power", 1000, 0, 1000);
     return property;
   }
-
 
   /**
    * This method returns a list of all supported resolutions (in DPI)
@@ -453,8 +455,10 @@ public class K40NanoDriver extends LaserCutter
     private String board = "M2";
     private double speed = 30;
     private int raster_step = 1;
+    private int power = 1000;
     double d_ratio = 0.2612;
 
+    private int power_remainder = 0;
     private int x = 0;
     private int y = 0;
 
@@ -497,6 +501,11 @@ public class K40NanoDriver extends LaserCutter
         exit_compact_mode();
       }
       speed = mm_per_second;
+    }
+
+    void setPower(int ppi)
+    {
+      power = ppi;
     }
 
     public int getRaster_step()
@@ -587,6 +596,7 @@ public class K40NanoDriver extends LaserCutter
 
     void cut_absolute(int x, int y)
     {
+
       int dx = x - this.x;
       int dy = y - this.y;
       cut_relative(dx, dy);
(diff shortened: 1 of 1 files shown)


tatarize:
Still needs a direct check on a proper machine but just changed the line draw algorithm to do carry-forward and it worked in mock mode.

t-oster:
So I should wait before merging, or do we just see if anyone complains?

tatarize:
@t-oster Wait to merge, it needs to have a physical test done on the actual hardware. But, for some reason I can't test it because it refuses to work. 

A fresh copy of the latest Netbeans the latests Visicut pull and even clean vs. new LibLaserCut and I can't launch.

```
Exit code: 1 - D:\Users\Tat\Documents\NetBeansProjects\VisiCut-0129\src\main\java\de\thomas_oster\uicomponents\AncorPointPanel.java:33: error: malformed HTML
* @author Thomas Oster <thomas.oster@rwth-aachen.de>
```

While the above code passed the tests I gave it for the mock driver it needs to be tested on a physical machine. I don't seem to be able to launch it. I also tried putting the jar in the nightly build and it didn't do more than sit there. Maybe an older nightly and replace the jar in that?

tatarize:
Cute, it thinks your email address in <> tags is a html tag in javadoc. So I need to close them.

 * @author Thomas Oster <thomas.oster@rwth-aachen.de>

Becomes:

 * @author Thomas Oster <thomas.oster@rwth-aachen.de/>

Oy, I just need a quick test to see that it still rasters and vectors work and the power setting exists and cuts at a lower power.

t-oster:
I always use '(' for E-Mail Adresses because javadoc can't read the '<'s.

tatarize:
Even bypassing that stuff it still never loads up a gui. Hm.

tatarize:
Loading platform independent version at least gave me an error. A JNI error.

"Error: A JNI error has occurred, please check your installation and try again."  At least this makes sense now. Maybe my broken JNA stuff is related. Shoulda got most of that write for the driver but it died before ever loading any dll ever.

tatarize:
@t-oster 
Maybe my mojo is gone but I can't compile it, like at all. It compiles and then tries to run but fails:

```
Downloaded: https://repo.maven.apache.org/maven2/org/sonatype/plexus/plexus-sec-dispatcher/1.3/plexus-sec-dispatcher-1.3.jar (28 KB at 44.6 KB/sec)
WARNING: An illegal reflective access operation has occurred
WARNING: Illegal reflective access by com.thoughtworks.xstream.core.util.Fields (file:/C:/Users/Tat/.m2/repository/com/thoughtworks/xstream/xstream/1.4.11.1/xstream-1.4.11.1.jar) to field java.util.TreeMap.comparator
WARNING: Please consider reporting this to the maintainers of com.thoughtworks.xstream.core.util.Fields
WARNING: Use --illegal-access=warn to enable warnings of further illegal reflective access operations
WARNING: All illegal access operations will be denied in a future release
Security framework of XStream not initialized, XStream is probably vulnerable.
Security framework of XStream not initialized, XStream is probably vulnerable.
Security framework of XStream not initialized, XStream is probably vulnerable.
```
Then it hangs... forever.

Could you or maybe somebody else, make a copy of visicut in it with that PR and let me run that. Or, test whether it seems to have added power, and if it does then go ahead and build it as a nightly and I'll use that for testing. I can seem to load the compiled vanilla nightlies. If I identify any issue it would only be a day or so turn around. Though power goes from 0 to 1000 since it's ppi. Not that scaling it to 100 would be hard.

t-oster:
Did you try to run the debugger and check where it hangs? I am not on a PC before Monday

tatarize:
```
    if (port == null) {
        port = ApplicationInstanceManager.getSingleInstancePort();
    }
```

That process hangs forever.  Calls readAllBytes, readNBytes, read, read1, fill, read, readBytes.

And readBytes is blocking.
```
    public int read(byte b[]) throws IOException {
        return readBytes(b, 0, b.length);
    }
```

tatarize:
Made that line:
```
            String sessionId = "";
//            String sessionId = new String(exec.getInputStream().readAllBytes()).strip();
```

And it loaded all the way.

tatarize:
It is throwing a warning while in Mock:

java.lang.Exception: Warning: A GUI function was called from the non-GUI thread Thread-4. This may cause sporadic errors and should therefore be fixed.
	at de.thomas_oster.visicut.gui.ThreadUtils.assertInGUIThread(ThreadUtils.java:70)
	at de.thomas_oster.visicut.gui.MainView$61$1.taskChanged(MainView.java:2212)
	at de.thomas_oster.liblasercut.drivers.K40NanoDriver.sendJob(K40NanoDriver.java:216)
	at de.thomas_oster.visicut.VisicutModel.sendJob(VisicutModel.java:793)
	at de.thomas_oster.visicut.gui.MainView$61.run(MainView.java:2261)

Which is because:
`this.progress.taskChanged(this, "Vector Part");`
Is called.

Also, pretty sure I set the power and it didn't trigger. I'll have it fixed after a bit. I can test now.

t-oster:
> Made that line:
> 
> ```
>             String sessionId = "";
> //            String sessionId = new String(exec.getInputStream().readAllBytes()).strip();
> ```
> 
> And it loaded all the way.

Do you have any process listen on the singleinstance port?

tatarize:
I dunno what that feature is or what it does. I do know it sent a blocking read request for data I didn't have on a vanilla fresh install of netbeans, visicut, and libLaserCut.


-- 

PPI works now, but I'll likely be able to to the required testing on a physical device later today or tomorrow. But, got it to the point where it *should* pass that testing.

Also, is that thing for real, where I'm given a progress listener and then warned that I shouldn't have touched it when I change the progress listener? I tossed the big changing on into the GUI thread but seems like that's really something that should be wrapped up and passed to the driver. A way to update the progress that doesn't yell at you if you use it.

t-oster:
The thing with the request is for assuring there is only one visicut instance running, either you have something listening on the same port by chance or we have to fix some timeouts.

The warning is from a relatively new commit because we had threadsafety issues in the GUI code. There is a one liner to execute in the correct thread, but you are right the progress listener should include that call. Will look into that tomorrow when I have time

tatarize:
Okay, if that's going to go that way, I'll get rid of my swingutilities gui thread call. Since it will be settled.  Also, there's no way to display any realtime information. I know all the other drivers don't provide that stuff since it's just send that data and you're finished. But, the Lhystudio boards and the old Moshi boards run differently and as such, need to be constantly spoon fed the data. My thread for the send hangs out for a while doing that. And really dies when you close the program without a how-do-you-do. And also provides a message that is utterly unhelpful and confusing to people from the report I got of somebody trying the driver.

I have nothing that runs on port 6543. Checked my netstat.

mgmax:
> And also provides a message that is utterly unhelpful and confusing to people from the report I got of somebody trying the driver.

I assume you are referring to the "job was sent successfully" message of VisiCut. This is configurable in Options - Lasercutters - Job sent text. You are right that the current default doesn't make sense with lasercutters that have no "print queue" but directly start the job. You are welcome to contribute a solution. Maybe we could just shorten the default to "Job xy sent successfully"?

mgmax:
Regarding the startup issue:
- Can you reproduce the hang with any official VisiCut build, or only with your custom one?
- Does the "powershell ..." call print something and exit correctly if you run it from "cmd"?
- Which Windows version are you using? Can you reproduce it on another PC or in a clean VM? (You can get Win10 test VMs from Microsoft here: https://developer.microsoft.com/windows/downloads/virtual-machines )

tatarize:
I have the bug disabled on in my current version. So steps I need to recreate it:

* Download latest Visicut snapshot.
* Load latest snapshot in Apache NetBeans IDE 11.2
* Click "Clean and Build."
* Click "Run"

Previously I had to manually install the three kabeja's packages but that was a quirk with the  maven repository. 

```
WARNING: An illegal reflective access operation has occurred
WARNING: Illegal reflective access by com.thoughtworks.xstream.core.util.Fields (file:/C:/Users/Tat/.m2/repository/com/thoughtworks/xstream/xstream/1.4.11.1/xstream-1.4.11.1.jar) to field java.util.TreeMap.comparator
WARNING: Please consider reporting this to the maintainers of com.thoughtworks.xstream.core.util.Fields
WARNING: Use --illegal-access=warn to enable warnings of further illegal reflective access operations
WARNING: All illegal access operations will be denied in a future release
Security framework of XStream not initialized, XStream is probably vulnerable.
Security framework of XStream not initialized, XStream is probably vulnerable.
---hang ---
```

tatarize:
Also, on the PPI front. The handle refused to attach even with the right driver. Might be something screwy with my java since it's failed other places it should not have failed. But, still no provable test of it totally working... yet.

mgmax:
I'm happy to merge it as soon as it has been successfully tested. Better not before.

Regarding the "startup hang" issue: Can you reproduce it in a clean VM?

tatarize:
Seems to reproduce that way. I went to my Appdata/Roaming directory and renamed the netbeans settings to kill all the settings I had there. And shutdown all my various programs. In my now fresh version of netbeans I ran a fresh pull of Visicut and it still hangs.

tatarize:
Seems like it you might not be able to easily reproduce it. The following code also hangs forever:

```java
public class Main {
    public static void main(String[] args) throws IOException {
        Process exec = Runtime.getRuntime().exec("powershell -Command (get-process -pid $pid).sessionid");
        String sessionId = new String(exec.getInputStream().readAllBytes()).strip();
    }
}
```

>powershell -Command (get-process -pid $pid).sessionid
>1

tatarize:
Read all bytes calls.
ReadNBytes(Integer.Max_VALUE)
Makes an 8k buffer to read bytes.
Calls read in BufferedInputStream to fill 8k buffer.
Calls read1 for the buffer.
returns  3 bytes. `1\r\n`
returns read.
Calls read again. Stream is still open.
Calls read1 again. characters available == 0
Calls fill, which calls a different read.  Which calls FileInpuStream readBytes() which is blocking. And will sit there until there's more bytes which there will never be since it already read all 3 bytes of the response.

tatarize:
Actually if I find the powershell running in the background and physically kill it, it will close the channel and return correctly.

tatarize:
```java
        Process exec = Runtime.getRuntime().exec("powershell -Command (get-process -pid $pid).sessionid");
        byte[] buffer = new byte[1024];
        int n = 0;
        while (true) {
            byte b = (byte)exec.getInputStream().read();
            if (b == -1 || b == 13) break;
            buffer[n++] = b;
        }
        String sessionId = new String(buffer).strip();
```

That code there works for fairly obvious reasons.

tatarize:
Though that keeps giving me `"1` which has a number format exception so:
```java
        Process exec = Runtime.getRuntime().exec("powershell -Command (get-process -pid $pid).sessionid");
        byte[] buffer = new byte[1024];
        int n = 0;
        while (true)
        {
          byte b = (byte) exec.getInputStream().read();
          if (b == -1)
          {
            break;
          }
          buffer[n++] = b;
          if (b == 10) {
            break;
          }
        }
        buffer = Arrays.copyOf(buffer, n);
        String sessionId = new String(buffer).strip();
```

Actually worked pasted in there. Though whatever code variations on the same theme. Just get 1 line of response then the failure to close the process won't live lock the stuff.

mgmax:
As it does not occur on recent Windows 10 1909, it is probably a bug in old powershell versions if you don't connect stdin: https://serverfault.com/questions/437504/why-does-psexec-hang-after-successfully-running-a-powershell-script
Can you try if the hang is gone if you call `cmd /c "echo exit | powershell -Command (get-process -pid $pid).sessionid;exit"`  ?
For me, the output is only `1`, not `"1`.

tatarize:
```
Exception in thread "main" java.io.IOException: Cannot run program "echo": CreateProcess error=2, The system cannot find the file specified
	at java.base/java.lang.ProcessBuilder.start(ProcessBuilder.java:1128)
	at java.base/java.lang.ProcessBuilder.start(ProcessBuilder.java:1071)
	at java.base/java.lang.Runtime.exec(Runtime.java:591)
	at java.base/java.lang.Runtime.exec(Runtime.java:415)
	at java.base/java.lang.Runtime.exec(Runtime.java:312)
	at com.mycompany.powershellreturn.Main.main(Main.java:16)
Caused by: java.io.IOException: CreateProcess error=2, The system cannot find the file specified
	at java.base/java.lang.ProcessImpl.create(Native Method)
	at java.base/java.lang.ProcessImpl.<init>(ProcessImpl.java:420)
	at java.base/java.lang.ProcessImpl.start(ProcessImpl.java:151)
	at java.base/java.lang.ProcessBuilder.start(ProcessBuilder.java:1107)
	... 5 more
```

The output should be 1. Not sure why it wasn't. Only read a buffer that said 1\x00\x00... etc. but, I truncated it having the carriage return and line feed and then clipped it to size.

mgmax:
Please leave in the "cmd /c".

tatarize:
Hm. I did omit that.

```java
public class Main {
    public static void main(String[] args) throws IOException {
        Process exec = Runtime.getRuntime().exec("cmd /c \"echo exit | powershell -Command (get-process -pid $pid).sessionid;exit\"");
        String sessionId = new String(exec.getInputStream().readAllBytes()).strip();
        System.out.println(sessionId);
    }
}
```

Gives:


```
Building powershellreturn 1.0-SNAPSHOT
------------------------------------------------------------------------

--- exec-maven-plugin:1.5.0:exec (default-cli) @ powershellreturn ---
1
------------------------------------------------------------------------
BUILD SUCCESS
------------------------------------------------------------------------
Total time: 1.356 s
Finished at: 2020-03-01T02:36:09-08:00
Final Memory: 7M/27M
------------------------------------------------------------------------

```

tatarize:
Backporting that into ApplicationInstanceManager.java causes it to launch without hang.

line 153:
`Process exec = Runtime.getRuntime().exec("cmd /c \"echo exit | powershell -Command (get-process -pid $pid).sessionid;exit\"");
`

mgmax:
fixed with ebf87835. Now we can get back to topic :-)

tatarize:
Okay, pretty sure the holdup is actually:
https://stackoverflow.com/questions/26012951/usb4java-usb-error-4-unable-to-open-usb-device

I too copied that code fragment from the usb4java page and it really is killing the device before I (quite literally) get a handle on it. *fingers cross* -- To the testing machine!

tatarize:
Stopped the hard crash at usb handle. Still couldn't test since it claimed I didn't have proper permissions. Which I now realize is because I didn't have the LibUsb driver installed. MeerK40t was just using windows driver. *sigh* I'll give it another shot.

tatarize:
I successfully tested this and had it complete a job using PPI. I now fully approve merging this.

The two Codacy/PR quality review items are non-appliquable.

I didn't manage to get my windows machine to run it but previous version was tested on Raspberry Pi and this one was similarly tested. Also, I corrected the main issue everybody reported for windows users. So if it failed previously in Windows it *should* be more likely to work. But, if it worked in Windows before it will not acquire new fail conditions.

t-oster:
If you'd remove the unused import and renamed the method to CamelCase Codacy would be happy

tatarize:
"Oops! This file no longer exist in the lastest commit of this branch."
"Oops! This file no longer exist in the lastest commit of this branch."

The unused import isn't a thing. The snake_case isn't a thing either. Also, they misspelled `latest`. 

See for yourself: https://github.com/K40Nano/LibLaserCut/blob/k40nano-ppi/src/main/java/de/thomas_oster/liblasercut/drivers/K40NanoDriver.java  There's no import for BlackWhiteRaster and there's nothing called raster_*

tatarize:
Lemme try adding a unneeded space and see if it refreshes.

tatarize:
Other test didn't seem to kick in after the bogus commit. But, it's objections were not thing that existed.

PR #139, opened 2020-02-05
Fix b/w engraving in convertRasterizableToVectorPart

mgmax:
All pixels inside the "bounding area" were engraved as if they were black.
Was maybe caused introduced by 0238bd1023cbc71fed9ae7901ba52fb47dc96186

Should fix https://github.com/t-oster/VisiCut/issues/562

Needs review - not properly tested, but seems to work at first look

### DIFF
--- a/src/main/java/de/thomas_oster/liblasercut/BlackWhiteRaster.java
+++ b/src/main/java/de/thomas_oster/liblasercut/BlackWhiteRaster.java
@@ -236,13 +236,13 @@ public class BlackWhiteRaster extends TimeIntensiveOperation implements Greyscal
   @Override
   public int getGreyScale(int x, int y)
   {
-    return getPixel(x, y);
+    return getPixel(x, y) * 255;
   }
 
   @Override
   public void setGreyScale(int x, int y, int grey)
   {
-    setPixel(x, y, grey);
+    setPixel(x, y, grey > 127 ? 1 : 0);
   }
 
 }
--- a/src/main/java/de/thomas_oster/liblasercut/BlackWhiteRaster.java
+++ b/src/main/java/de/thomas_oster/liblasercut/BlackWhiteRaster.java
@@ -236,13 +236,12 @@ public class BlackWhiteRaster extends TimeIntensiveOperation implements Greyscal
   @Override
   public int getGreyScale(int x, int y)
   {
-    return getPixel(x, y) * 255;
+    return isBlack(x,y) ? 0 : 255;
   }
 
   @Override
   public void setGreyScale(int x, int y, int grey)
   {
-    setPixel(x, y, grey > 127 ? 1 : 0);
+    setBlack(x, y, grey < 127);
   }
-
 }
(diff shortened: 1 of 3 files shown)


pimsierhuis:
We have this problem too, but this PR gives us an inverted image.

The following code (BlackWhiteRaster.java) yielded better results for us:

`
public int getGreyScale(int x, int y)
{
  if (isBlack(x,y)) {
    return 0;
  } else {
    return 255;
  }
}

public void setGreyScale(int x, int y, int grey)
{
  setBlack(x, y, grey < 127);
}

`

t-oster:
@tatarize you broke it, you fix it? ;)

tatarize:
The problem here is likely due to the weird non-conformance issue. In black and white, 1 counts as black. But, in 8 bit grey, 0 counts as black. You can't just scale it exactly because it also needs to get flipped weirdly. If 0-1 is white and black and you're making it 0-255 with black to white. It not only scales up by 255 but is 255 - (c * 255). Lemme doublecheck.

tatarize:
It looks like pimsierhuis's code there is going to be right. getGreyScale() is wrongly returning the raw data in blackwhite raster.  The expected value is actually a value between 0 and 255 since it parallels the code for the buffer image adapter. That the 1 bit rasters are inverted just means that mg's initial fix missed the mark. Since he scaled by 255. which ignored that other conformance factor. 

```java
public int getGreyScale(int x, int y)
{
   return isBlack(x,y)? 0: 255;
}

public void setGreyScale(int x, int y, int grey)
{
  setBlack(x, y, grey < 127);
}

```

This should be strongly preferred over mg's code fix there since it delegates to isBlack() properly solves for the conformance issues between the data meaning.

mgmax:
I added @tatarize's fix and some regression tests. Can someone affected by the bug please test if it is fixed?

tatarize:
In conclusion, yes. I broke it. The danger in the switch to the delegate was that if it wasn't flawless it could cause a flaw. And it wasn't flawless. The expectation for grey was a range of 0-255 within the expected 0=black 255=white greyscale range, which required the same conformance elements as isBlack used.

pimsierhuis:
As that code is equivalent to what i commented (doublechecked) and tested my code on our cutter, i'd say it's good.

tatarize:
[visicut 1.zip](https://github.com/t-oster/LibLaserCut/files/4206441/visicut.1.zip)
Here's the original mario, smoothie board, with 50% power and the fix:

```gcode
G1 X31.648400 Y4.114800 S0.000000
G1 X31.699200 Y4.114800 S0.500000
G1 X31.800800 Y4.114800 S0.000000
G1 X31.851600 Y4.114800 S0.500000
G1 X31.953200 Y4.114800 S0.000000
G1 X32.004000 Y4.114800 S0.500000
G1 X32.105600 Y4.114800 S0.000000
G1 X32.156400 Y4.114800 S0.500000
G1 X32.258000 Y4.114800 S0.000000
G1 X32.308800 Y4.114800 S0.500000
G1 X32.410400 Y4.114800 S0.000000
G1 X32.461200 Y4.114800 S0.500000
G1 X32.562800 Y4.114800 S0.000000
G1 X32.613600 Y4.114800 S0.500000
G1 X32.715200 Y4.114800 S0.000000
```

Basically all looks like this.

PR #141, opened 2020-02-19
Semi-automated cleanup of a few files

phdoerfler:


t-oster:
Looks correct as a first glance, what cleaner/IDE did you use? Not sure if empty @param comments should be left out and if interface methods are public by default and if abstract methods do not need to be declared abstract anymore, but I guess you tested it?

src/main/java/de/thomas_oster/liblasercut/GreyscaleRaster.java
```diff
@@ -24,28 +24,20 @@
...
 
-  public int getWidth();
+  int getWidth();
 
-  /**
-   * get greyscale value of pixel
-   * @return greyscale value: 0 (black) ... 255 (white)
-   */
-  public int getGreyScale(int x, int y);
+  int getGreyScale(int x, int y);
```
mgmax: you accidentally removed a comment.

src/main/java/de/thomas_oster/liblasercut/GreyscaleRaster.java
```diff
@@ -24,28 +24,20 @@
...
 
-  /**
-   * set greyscale value of pixel
-   * @param grey greyscale value: 0 (black) ... 255 (white) 
-   */
-  public void setGreyScale(int x, int y, int grey);
+  void setGreyScale(int x, int y, int grey);
```
mgmax: same here: you accidentally removed a comment. It's important to document which value is black and which is white, there were confusions in the past.

src/main/java/de/thomas_oster/liblasercut/LaserProperty.java
```diff
@@ -34,16 +34,16 @@
...
   
-  public abstract Object[] getPossibleValues(String name);
+  Object[] getPossibleValues(String name);
   
-  public abstract LaserProperty clone();
+  LaserProperty clone();
   
   // Please override equals so that it works well
   // Otherwise there is trouble in the GUI when it tries to compare laser settings
   @Override
```
mgmax: Does "@Override" make any sense here?

phdoerfler:
Good catch! That was indeed unintentional and slipped in during the merge

src/main/java/de/thomas_oster/liblasercut/LaserProperty.java
phdoerfler: Yes and no. Yes in that it correctly indicates that Object.equals is overriden.
No because overriding Object.equals in an interface makes little sense, especially given that hashCode is not overriden. The contract of Object.equals and Object.hashCode demand that you deal with both.
Overriding equals in an interface makes sense when you can narrow down the type signature, but as it is here it accepts any Object, so it provides no benefit over the original Object.equals.
Overriding equals in an interface does not guarantee that it is overridden in implementing classes.
I am afraid that this is beyond Java's type system.

mgmax:
Thanks, I will test and merge this in the next days together with other changes.

PR #145, opened 2020-03-08
Automated test that driver output doesn't change

mgmax:
@phdoerfler to avoid duplicate work, here's a quick experiment I did before your reply


mgmax:
Should now be good enough for merging - any feedback?

t-oster:
Looks good on first sight... I suppose it cannot do any harm, so there's nothing against merging

PR #148, opened 2020-03-14
Deduplicate and extend pseudo-raster engrave functions

### LINKED ISSUE #138: convertRasterizableToVectorPart overscan may exceed the laser bed
`convertRasterizableToVectorPart()` ignores the minimum and maximum x values of the lasercutter.
This is a problem if an engraving goes all the way to the left or right border and the raster overscan is nonzero.
We need a parameter for the minimum and maximum x. Then overscan moves should be limited to that. (Note that for lasercutters using relative coordinates to handle custom startpoints, these limits must be changed if a startpoint is used.)

mgmax:
Fixes #138: Raster overscan may exceed the laser bed

and removes hundreds of lines of duplicate code.

The output of many drivers changed, testing engrave and engrave3d would be good:
- GenericGCode: looks good, almost unchanged
- GoldCutHPGL: importing the output to Inkscape looks good
- [ ] Laos: engrave3d needs testing, power values seem to have changed slightly
- [ ] Lasersaur: unsure
- [ ] MakeBlockXYPlotter: unsure
- Marlin: looks good
- Smoothieboard: looks good


tatarize:
Can you also look into the RasterBuilder class. The idea there was/is to do the rastering in a purely agnostic fashion, with easy editing and state-based mode changes. It should be able to give a centralized class that does rastering and can alter the way the rasters work without altering how they are implemented for the drivers.

mgmax:
The current job was just to collect everything that is more or less a copy of convertRasterizableToVectorPart (plus a few bugs in each of the different variants). I think this is a first step that should be done even if it is later rewritten using RasterBuilder or some other iterator-like scheme that avoids the overhead of storing the VectorPart and then parsing it again.

mgmax:
I merged although Laos, Lasersaur and MakeBlockXYCutter are untested, because I didn't want to get stuck waiting for a response, and because this change also fixes important issues (machine limits not respected with overscan or startpoint). In the future it would be cool to have a list of beta testers for each device that can help by testing jobs on a real machine. Or at least complete the list of supported devices https://github.com/t-oster/VisiCut/wiki/Supported-Hardware .

PR #149, opened 2020-03-30
Code Cleanup assisted by IntelliJ IDEA

phdoerfler:
Replaced invocations of deprecated API with their new counterparts, converted anonymous classes to lambdas, added a few type parameters, used List's sort instead of Collections.sort, removed superfluous generic type parameters, used Double.compare instead of creating new Double instances and comparing those (which internally uses Double.compare anyway), replaced for loops with Java 5 enhanced for loops, used resource aware try.

### DIFF
--- a/src/main/java/de/thomas_oster/liblasercut/AbstractLaserProperty.java
+++ b/src/main/java/de/thomas_oster/liblasercut/AbstractLaserProperty.java
@@ -38,7 +38,7 @@ public class AbstractLaserProperty implements LaserProperty
   static final String MAX_SUFFIX = "_maximum";
   static final String MIN_SUFFIX = "_minimum";
   static final String VALUES_SUFFIX = "_values";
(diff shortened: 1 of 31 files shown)


mgmax:
There are a few conflicts, could you please rebase? Then I'll merge quickly.

phdoerfler:
should be done :)

phdoerfler:
> should be done :)

I knew I shouldn't have said that

mgmax:
Looks like you accidentally restored some functions that were deleted.

phdoerfler:
> Looks like you accidentally restored some functions that were deleted.

Ah that would explain it. I think my merge tool showed the local as remote changes and vice versa so to me it looked like those functions were added.

phdoerfler:
Apparently during a rebase git on my machine does indeed label my changesets as "remote" and the incoming changesets as "local". I smell leaky abstraction.

mgmax:
thanks :-)

PR #150, opened 2020-04-05
Join broken paths on inner-first sorting with tolerance

mgmax:
In some files, paths are chopped into pieces, which confuses the INNER_FIRST algorithm. Previously, INNER_FIRST only merged exactly identical start/end points, which is problematic: In floating point, ((a + b) + c) == ((a + b) + c) isn't always true. For example, in files from CAD ( [example](https://www.thingiverse.com/download:1984102) ) or in formats that mix absolute and relative coordinates (importing CAD files to Inkscape, see [problems.zip](https://github.com/t-oster/LibLaserCut/files/4434731/problems.zip)). 

The problem got a lot worse since Point in LibLaserCut was changed from int to double (which was required for post-processing in the LaserToolsTechnics driver). Before, most 

My proposed change introduces a tolerance of 0.9 pixels (in the current DPI), so about 0.02mm for 1000dpi. Fuzzy comparison comes at the cost of CPU time: INNER_FIRST sorting is now significantly slower than the old algorithm from #84 , but only about 2-5x slower than NEAREST (which has quadratic complexity).

The example file has 20.000 paths which are joined into 380. This takes about 10 seconds to process, compared to 3 seconds for NEAREST. For comparison, many operations in Inkscape on this file are slower. 
The time grows roughly quadratic with the number of paths: The worst case I tested is 80.000 paths. Then, the whole VisiCut UI is already slow but the algorithm still more or less works (about 180 seconds for INNER_FIRST, compared to 60 seconds for NEAREST).
This is okay-ish but still much better than before if you consider that it saves lots of frustration and laser time.

There is some room for optimization because currently every path is compared to almost every other one, while actually we could sort the elements by x-position and then only compare elements with similar x-coordinates. Implementing this will be some fun because we have to keep the list sorted while modifying elements.

As a side note, I don't understand why NearestVectorOptimizer is fastest with its current choice of ArrayList than with any other possibility I tried (especially ArrayList with setting to null instead of removing, and also LinkedList). Heavily using .remove() shouldn't be the best idea for ArrayList...

### DIFF
--- a/src/main/java/de/thomas_oster/liblasercut/platform/Point.java
+++ b/src/main/java/de/thomas_oster/liblasercut/platform/Point.java
@@ -84,7 +84,7 @@ public class Point
    * compute euclidean distance to another point
    * @param p other point
    */
-  public double hypothenuseTo(Point p) {
+  public double hypotTo(Point p) {
     return this.subtract(p).hypot();
   }
   
--- a/src/main/java/de/thomas_oster/liblasercut/platform/Point.java
+++ b/src/main/java/de/thomas_oster/liblasercut/platform/Point.java
@@ -80,21 +80,21 @@ public class Point
     return "Point(" + x + ", " + y + ")";
   }
 
-  public int compareTo(Point o)
-  {
-    if (y < o.y) return -1;
-    if (y > o.y) return 1;
-    if (x < o.x) return -1;
-    if (x > o.x) return 1;
-    return 0;
-  }
-
     /**
    * compute euclidean distance to another point
    * @param p other point
    */
   public double hypothenuseTo(Point p) {
     return this.subtract(p).hypot();
+  }
+  
+  /**
+   * Compute Manhattan (1-norm) distance to another point.
+   * Fast approximation of euclidean distance.
+   * @return |self.x - p.x| + |self.y - p.y|
+   */
+  public double manhattanDistanceTo(Point p) {
+    return this.subtract(p).manhattanLength();
   }
 
   public Point add(Point p) {
@@ -135,6 +135,15 @@ public class Point
    */
   public double hypot() {
     return Math.sqrt(x*x + y*y);
+  }
+  
+  /**
+   * Get manhattan (1-norm) length.
+   * Useful as fast approximation of hypot()
+   * @return  |x| + |y|
+   */
+  public double manhattanLength() {
+    return Math.abs(x) + Math.abs(y);
   }
 
   /**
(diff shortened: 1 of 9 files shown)


mgmax:
With some optimizations it's now fast enough for 80k paths (few seconds, compared to 50 seconds for NEAREST). At this point the GUI becomes the major bottleneck (clicking anything take seconds). If there's no feedback in the next time, I'll merge it.

@Hueburtsonly @rbowmaker any feedback?

src/main/java/de/thomas_oster/liblasercut/vectoroptimizers/OptimizerUtils.java
```diff
@@ -20,162 +20,291 @@
...
+              // For every candidate point, check if it's nearby.
+              // If yes, remember it in startNearCurrentHead or endNearCurrentHead.
+              if (e == null || e.index == current.index)
               {
-                i++;
+                // skip removed (null) entries,
+                // skip current element
                 continue;
               }
+              // skipUntil guarantees that all points are not too far left of the current head point.
```
rbowmaker, line 149: skipPoints

src/main/java/de/thomas_oster/liblasercut/vectoroptimizers/OptimizerUtils.java
```diff
@@ -20,162 +20,291 @@
...
+              // Nothing to do. After the start point (invert==0),
+              // the same will be checked for the end point (invert==1)
+              // and then for the next point (i++).
+            }
+          }
+          if (!hasAnyNeighbors)
           {
-            break;
+            // neither start nor end have any points nearby.
+            // -> information about this path is not relevant for remaing paths
```
rbowmaker, line 273: remaining

t-oster:
> At this point the GUI becomes the major bottleneck (clicking anything take seconds). 
We probably could implement some preview like with the dithering algorithms and wait for the calculation in a thread? With the christmas-tree I could not see any big difference in lagging thogh.

src/main/java/de/thomas_oster/liblasercut/platform/Point.java
```diff
@@ -80,22 +80,22 @@ public String toString()
...
    * @param p other point
    */
   public double hypothenuseTo(Point p) {
     return this.subtract(p).hypot();
   }
+  
+  /**
+   * Compute Manhattan (1-norm) distance to another point.
+   * Fast approximation of euclidean distance.
+   * @return |self.x - p.x| + |self.y - p.y|
```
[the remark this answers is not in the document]
mgmax: I can't really tell, time measurements with the java JIT are a bit random. Feel free to find out ;-)

mgmax:
@t-oster the lag in the GUI is probably due to rendering the paths. This is not cached yet, although it could be in the future (as soon as the other engrave caching stuff is merged in).

PR #151, opened 2020-04-09
Fixed dangling Javadoc comments

phdoerfler:
Having Javadoc at the beginning of a .java file is an error. I converted these into block comments.

### DIFF
--- a/copyrightheader
+++ b/copyrightheader
@@ -1,18 +1,18 @@
-/**
- * This file is part of LibLaserCut.
- * Copyright (C) 2011 - 2014 Thomas Oster <mail@thomas-oster.de>
- *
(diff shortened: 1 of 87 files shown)


t-oster:
We have to update the test.sh/COPYRIGHTHEADER also

phdoerfler:
/sighs/ I did run test.sh locally and it didn't complain.

phdoerfler:
this is a pain, I wish we were using https://github.com/sbt/sbt-header or whatever equivalent there is in maven land

phdoerfler:
Dunno why test.sh did not complain before. I guess I must have run it on the visicut repository accidentally. Anyways, it should™ be fixed now.

t-oster:
Feel free to migrate it to whatever you think fits better

PR #153, opened 2020-04-23
Fixed a couple of probable bugs and cleaned up some more code

phdoerfler:
I am very sceptical of Tuple equals. Are we sure this fullfills the contract dictated by Object.equals esp. in respect to hashcode?

### DIFF
--- a/src/main/java/de/thomas_oster/liblasercut/drivers/EpilogCutter.java
+++ b/src/main/java/de/thomas_oster/liblasercut/drivers/EpilogCutter.java
@@ -446,7 +446,7 @@ abstract class EpilogCutter extends LaserCutter
     {
       int p;
       p = idx + 1;
-      while (p < r && p < idx + 128 && line.get(p) == line.get(idx))
+      while (p < r && p < idx + 128 && line.get(p).equals(line.get(idx)))
       {
         p++;
       }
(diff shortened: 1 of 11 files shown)


src/main/java/de/thomas_oster/liblasercut/drivers/EpilogCutter.java
```diff
@@ -446,7 +446,7 @@ public void encode(List<Byte> line, List<Byte> result)
     {
       int p;
       p = idx + 1;
-      while (p < r && p < idx + 128 && line.get(p) == line.get(idx))
+      while (p < r && p < idx + 128 && line.get(p).equals(line.get(idx)))
```
mgmax, line 449: Surprisingly, this does not change the driver output. Do you have any idea why the old version worked?

t-oster: Maybe some Java boxing/unboxing magic?

phdoerfler: No magic involved. It's all in the docs: https://docs.oracle.com/en/java/javase/13/docs/api/java.base/java/lang/Integer.html#valueOf(int).
Autoboxing uses `#valueOf` of Integer, Byte, Char, etc. which by default caches some values (e.g. from -128 to +127). Invocations of `#valueOf` with these numbers will return the cached instances which in turn means that a reference comparison with `==` will work as "expected" for those values.
The size of this cache used to be hardcoded but can nowadays be adjusted with a system property.
Needless to say that it is not a good idea to rely on this caching and when crossing the boundaries it can lead to some really hard to find bugs. I've been there and I don't want to go there ever again.

phdoerfler:
Codacy rightfully pointed out that $f in test.sh was not quoted yet. This is now fixed as well.

t-oster:
Since this won't affect the behavior, I will delay updating visicut

PR #156, opened 2020-05-12
Code cleanup

phdoerfler:
This one was largely removal of unused assignments. Nothing crazy.


mgmax:
At first look I thought you changed the behavior of the code at some points, but on second look the new code is actually easier to understand and still equivalent. Thanks!

PR #157, opened 2020-05-13
t-oster/VisiCut#577 Return 0,0, not Home

tatarize:
Correction for t-oster/VisiCut#577 rather than home returning to true 0,0 we return simply absolute move to 0,0 so if we began this operation at an adjusted position, we maintain that adjustment rather than lose it for subsequent operations.

### DIFF
--- a/src/main/java/de/thomas_oster/liblasercut/drivers/K40NanoDriver.java
+++ b/src/main/java/de/thomas_oster/liblasercut/drivers/K40NanoDriver.java
@@ -264,6 +264,7 @@ public class K40NanoDriver extends LaserCutter
         }
       }
     }
+    device.exit_compact_mode();
     device.move_absolute(0, 0); //Return device to start position 0,0.
     device.execute();
     device.close();
--- a/src/main/java/de/thomas_oster/liblasercut/drivers/K40NanoDriver.java
+++ b/src/main/java/de/thomas_oster/liblasercut/drivers/K40NanoDriver.java
@@ -264,7 +264,7 @@ public class K40NanoDriver extends LaserCutter
         }
       }
     }
-    device.home(); //Home the device after the job.
+    device.move_absolute(0, 0); //Return device to start position 0,0.
     device.execute();
     device.close();
   }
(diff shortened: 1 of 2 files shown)


tatarize:
The failed check is against the previously stated output. This correction modifies the operation from Home `IPP` to a move operation `IL(xxx)T(bbb)S1P` back to the start position through an absolute move. By definition this changes the job output.

t-oster:
This is actually consistent to the Epilog driver's behavior and better than homing IMHO. Please fix the expected driver output so I can merge

tatarize:
I'm not sure where the previous expected output is stored, or how that test actually works.

t-oster:
well, the test is pretty clear about that:
```
Output for de.thomas_oster.liblasercut.drivers.K40NanoDriver changed. For details, compare the files ./test-output/de.thomas_oster.liblasercut.drivers.K40NanoDriver.out and ./test-output/de.thomas_oster.liblasercut.drivers.K40NanoDriver.out.new manually. If this change is okay, delete the old file ./test-output/de.thomas_oster.liblasercut.drivers.K40NanoDriver.out , rerun the tests twice and then don't forget to commit the changed file.

tatarize:
The test output isn't what I sort of expected. There might be a random modal switch it's missing. I'll want to test this on a physical machine. Should be able to do so in the morning. I'll have to quickly switch out my Moshiboard and hook up the Pi, but it's pretty straight forward. Shouldn't be the lag time it was for the PPI upgrade.

tatarize:
Ah, there is. The home required the jump to default mode, but the absolute move does not require that. So I need to be explicit about that. Since I can't actually expect to safely perform that command inside the compact raster block with a step code (could go very badly).

tatarize:
By the way, that test fails in windows to boot since it has the windows vs. unix ending problem. Pretty sure the latest there will work. Actually gives me the output I expected. I'll give it a shot on my K40 tomorrow, with a moved laserhead, vector and raster.

tatarize:
@t-oster Yeah, it did exactly what was expected, when tested on the physical machine with multipart jobs and several different consecutive jobs.

PR #169, opened 2021-01-27
added driver for K3/K5/K3 Mini Laser engraver machines

mariolukas:
added support for K3/K5/K6 Mini Laser engravers.
this driver is only capable to engrave
it implements the reverse engineered protocol from https://github.com/RBEGamer/K3_LASER_ENGRAVER_PROTOCOL

### DIFF
--- a/src/main/java/de/thomas_oster/liblasercut/LibInfo.java
+++ b/src/main/java/de/thomas_oster/liblasercut/LibInfo.java
@@ -55,7 +55,8 @@ public class LibInfo
       Marlin.class,
       FullSpectrumCutter.class,
       LaserToolsTechnicsCutter.class,
-      K40NanoDriver.class
+      K40NanoDriver.class,
+      K3EngraverDriver.class
     );
   }
 }
--- a/src/main/java/de/thomas_oster/liblasercut/drivers/K3EngraverDriver.java
+++ b/src/main/java/de/thomas_oster/liblasercut/drivers/K3EngraverDriver.java
@@ -1,0 +1,891 @@
+/*
+ * To change this license header, choose License Headers in Project Properties.
+ * To change this template file, choose Tools | Templates
+ * and open the template in the editor.
+ */
+ /*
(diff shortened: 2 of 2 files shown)


mariolukas:
It seems that i forgot to implement saveJob.

t-oster:
yep, and also please make sure the copyright header is correct. You can run the test.sh locally to confirm.

t-oster:
but thanks for the PR and your effort, will merge as soon as the CI checks pass.

mgmax:
I'm not sure if the pull request was already ready to merge. There are several points that should be improved so that we have less effort maintaining the driver in the long term. For example, unused variables, commented-out code, unused imports, and probably extra variables in the VisiCut XML export. The latter is important becauses changes will break compatibility with existing settings files.

mariolukas:
You are right, i removed some unused import, variables and commented-out-code. But i don't understand what you mean with "extra variables in the VisiCut XML export".

t-oster:
Look at the xml file in your .visicut folder when you have configured the laser cutter and see if everything is necessary. The driver instance gets serialized to xml and if we remove properties of the class later that can lead to files not being compatible anymore

mgmax:
I added some details on the "XML export" issue to the Wiki page: https://github.com/t-oster/VisiCut/wiki/Developing-a-new-Lasercutter-Driver

mgmax:
I'm not happy that the driver was merged in before everything was ready. We need proper code quality so that drivers can be kept in the long term. In particular, fixing the XML export will break old settings version, so this should happen soon or we should temporarily remove the driver again.

@mariolukas Did the new infos on the Wiki page help? Do you need assistance?

t-oster:
@mgmax I partially agree. However a not perfect driver is better than no driver. If a future update breaks existing settings, I think we should also fix that problem e.g. by making the xml import more robust against such problems.

mgmax:
I have merged some code cleanup that was already in @mariolukas repository.

PR #175, opened 2021-12-04
Z-axis moves comply with FORCE LASER OFF DURING G0. Added settings for # of trailing digits.

McNugget6750:
The current build does not support Z-Axis moves to comply with "Force Laser Off During G0 Moves". This change adds this functionality as discussed in https://github.com/t-oster/VisiCut/issues/626.

Also added settings to allow configuration of desired number of trailing digits for gCode generator. My GRBL machine works a lot smoother with less digits and cannot resolve more than 3 digits realistically anyways.
And since my laser settings go from 0 - 700, there is no need for six trailing zeros to that number either. It can now be configured and the numbers get reformatted accordingly in move() and line(). Any additional trailing zeros are also removed in the process.


t-oster:
Hi. Thanks for your contribution. Could you please remove the commented code. Also I guess two instances of Decimal Format initialized on job start would be better than creating it everytime.
The version suggested by @mgmax with improved readability looks nicer to me, because it does not duplicate Code in both if branches.

McNugget6750:
Testing the changes now

McNugget6750:
Since the instance of DecimalFormat is not needed anywhere else at the moment, I'm hesitant to make it something like a global variable. It should not cost much to instantiate this. But I'm happy to change it to whatever you like. Open to suggestions.

t-oster:
Well a private instance variable is not really global. I would create two so you would not have to change the number of decimals on every G command but just before starting a job, but that may be a matter of taste.

If you want you can squash your commits and force push, so we get a single clean commit instead of three.

src/main/java/de/thomas_oster/liblasercut/drivers/GenericGcodeDriver.java
```diff
@@ -1057,6 +1073,28 @@ public void setApiKey(String apiKey)
     this.apiKey = apiKey;
   }
   
+  private int gCodeDigits;
```
mgmax, line 1076: is the default value correct? what happens when importing old settings?

McNugget6750:
I googled a bit and I'm still not sure how to accomplish this without messing things up on my end.

mgmax (changes requested):

src/main/java/de/thomas_oster/liblasercut/drivers/GenericGcodeDriver.java
McNugget6750: I have not set default values and I have not tested importing existing setting files. When I fired up the new build they initialized with 0.

mgmax:
Also, the regression tests currently fail (https://app.travis-ci.com/github/t-oster/LibLaserCut/builds/242917542#L719-L723), please follow the instructions there to recreate the files, and check if the differences are okay. If instructions are unclear then please ask so we can improve them :-)

McNugget6750:
I will read through them. I will say though, that those tests already failed when I cloned the fresh repository in the first place. Not sure what's the cause at the moment.

McNugget6750:
The regression test that fails on my end is not my code:
```
Generating test output for driver K3EngraverDriver
Your driver does not implement saveJob(LaserJob job)
Warning: Cannot test driver de.thomas_oster.liblasercut.drivers.K3EngraverDriver because it does not support saveJob()
Tests run: 1, Failures: 0, Errors: 1, Skipped: 0, Time elapsed: 0.361 sec <<< FAILURE!
```

McNugget6750:
Oh, I see it now. I need to look into this more:

`Output for de.thomas_oster.liblasercut.drivers.GenericGcodeDriver changed. For details, compare the files .\test-output\de.thomas_oster.liblasercut.drivers.GenericGcodeDriver.out and .\test-output\de.thomas_oster.liblasercut.drivers.GenericGcodeDriver.out.new manually. If this change is okay, delete the old file .\test-output\de.thomas_oster.liblasercut.drivers.GenericGcodeDriver.out , rerun the tests twice and then don't forget to commit the changed file.`

src/main/java/de/thomas_oster/liblasercut/drivers/GenericGcodeDriver.java
McNugget6750: I added default values and creating an new machine shows the appropriate values now. When importing an old configuration, these values will initialize with 0 and have to be set manually again. When importing settings that where exported AFTER this change, these values initialize with the values saved to the export as expected.

mgmax (changes requested):
Thanks for your effort! We're getting closer to a good result. I found some remaining issues that may need fixing.

In hindsight, it may have been better to split this up into two or three pull requests; one for the formatting (nice to have) and one for the Z-moves / S-values (critical). But that would be too much effort now.

src/main/java/de/thomas_oster/liblasercut/drivers/GenericGcodeDriver.java
```diff
@@ -494,45 +499,56 @@ protected void setSpeed(double speedInPercent) {
...
-    {
-      sendLine("G0 Z%f", focus);
-      currentFocus = focus;
+    DecimalFormat df = new DecimalFormat();
+    df.setMaximumFractionDigits(getGCodeDigits());
+    if (currentFocus != focus) {
+       String append = "";
+       if (blankLaserDuringRapids) {
+          append = " S0";
+          currentPower = 0.0;
```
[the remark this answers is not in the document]
McNugget6750: Did you test this? This had no effect when I tried this on my end - unless I didn't produce the correct setting configuration when I tried.

src/main/java/de/thomas_oster/liblasercut/drivers/GenericGcodeDriver.java
```diff
@@ -494,45 +499,56 @@ protected void setSpeed(double speedInPercent) {
...
   }
 
   protected void move(PrintStream out, double x, double y, double resolution) throws IOException {
     x = isFlipXaxis() ? getBedWidth() - Util.px2mm(x, resolution) : Util.px2mm(x, resolution);
     y = isFlipYaxis() ? getBedHeight() - Util.px2mm(y, resolution) : Util.px2mm(y, resolution);
     currentSpeed = getTravel_speed();
+    DecimalFormat df = new DecimalFormat();
     if (blankLaserDuringRapids)
     {
       currentPower = 0.0;
```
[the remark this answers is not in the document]
McNugget6750: Same here. I wasn't able to make this work they way you explained it might work. But I can test this again in more detail once I find the time.

test-output/de.thomas_oster.liblasercut.drivers.Grbl.out
```diff
@@ -1 +1 @@
...
G0X0.000000Y4.064000S0
G1X2.2352Y4.064F1200
G1X2.5908Y4.064S500
G1X7.5692Y4.064S0
G0X0.000000Y4.165600S0
G1X2.1336Y4.1656F1200
G1X2.2352Y4.1656S1,000
G1X7.2136Y4.1656S0
M5
G0X0Y0
```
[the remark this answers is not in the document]
McNugget6750: uh! This is a good catch! Yeah, that should NOT have a comma in there. I look into this.

src/main/java/de/thomas_oster/liblasercut/drivers/GenericGcodeDriver.java
```diff
@@ -494,45 +499,56 @@ protected void setSpeed(double speedInPercent) {
...
 
   protected void line(PrintStream out, double x, double y, double resolution) throws IOException {
     x = isFlipXaxis() ? getBedWidth() - Util.px2mm(x, resolution) : Util.px2mm(x, resolution);
     y = isFlipYaxis() ? getBedHeight() - Util.px2mm(y, resolution) : Util.px2mm(y, resolution);
     String append = "";
+    DecimalFormat df = new DecimalFormat();
     if (nextPower != currentPower)
     {
-      append += String.format(FORMAT_LOCALE, " S%f", nextPower);
+      df.setMaximumFractionDigits(getSCodeDigits());
```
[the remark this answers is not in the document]
McNugget6750: Makes sense. I changed it to unique objects. More on that below.

test-output/de.thomas_oster.liblasercut.drivers.Grbl.out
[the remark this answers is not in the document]
McNugget6750: Okey, I read up on this and extended the code. I agree that the DecimalFormat object seems cumbersome. However, I couldn't come up with a good way to make something like println("%3.3f"); in a println adjustable. The DecimalFormat object allows me to simply feed it an integer and it does the rest for me.
I added a locale and set it to en, US and then applied a numbering pattern that does not include a , or similar at the 1000s. The output looks clean now but the function got a bit longer. However, with the other cleanup you mentioned above, it's pretty readable now.

Throwing everything into a utility function might still make the most sense from a clarity point of view. Let me look at the util file you suggested.

McNugget6750:
I added the changes, reran the tests, all comes out clear. I committed my changes to my fork. The last couple times, the changes showed up here directly but not this time. Do I need to close this pull request and start a new one for this?

McNugget6750:
I apologize for the commit chaos all of a sudden. I was expecting to be able to create another pull request for the raster padding into negative space addition but I don't see how that works. I'll try and clean this up once I find a good solution. Any tips? It seems like I would need to create multiple branches to be able to do this. Man, I didn't think I'd need to become a git expert to contribute to a cool open source project... Git alone turns into an entire project all of a sudden...

src/main/java/de/thomas_oster/liblasercut/drivers/GenericGcodeDriver.java
```diff
@@ -494,45 +500,73 @@ protected void setSpeed(double speedInPercent) {
...
   protected void setFocus(PrintStream out, double focus) throws IOException {
-    if (currentFocus != focus)
-    {
-      sendLine("G0 Z%f", focus);
-      currentFocus = focus;
+    Locale locale  = new Locale("en", "US");
+    String formatPattern = "###.##";
+    DecimalFormat coordinateFormat = (DecimalFormat)NumberFormat.getNumberInstance(locale);
+    coordinateFormat.applyPattern(formatPattern);
+    coordinateFormat.setMaximumFractionDigits(getGCodeDigits());
```
mgmax, line 509: Can you please deduplicate this code block? It is present about four times; a separate function would be easier to read. For example `String formatFloat(float value, int decimalPlaces)`.

src/main/java/de/thomas_oster/liblasercut/drivers/GenericGcodeDriver.java
```diff
@@ -494,45 +500,73 @@ protected void setSpeed(double speedInPercent) {
...
+    String formatPattern = "###.##";
+
+    DecimalFormat coordinateFormat = (DecimalFormat)NumberFormat.getNumberInstance(locale);
+    coordinateFormat.applyPattern(formatPattern);
+    coordinateFormat.setMaximumFractionDigits(getGCodeDigits());
+
+    DecimalFormat sValueFormat = (DecimalFormat)NumberFormat.getNumberInstance(locale);
+    sValueFormat.applyPattern(formatPattern);
+    sValueFormat.setMaximumFractionDigits(getSCodeDigits());
+
```
mgmax, line 558: see above

src/main/java/de/thomas_oster/liblasercut/drivers/GenericGcodeDriver.java
```diff
@@ -1146,8 +1221,14 @@ public Object getProperty(String attribute) {
       return this.getUploadMethod();
     } else if (SETTING_RASTER_PADDING.equals(attribute)) {
       return this.getRasterPadding();
+    } else if (SETTING_RASTER_PADDING_ALLOW_NEGATIVE_SPACE.equals(attribute)) {
+      return this.getRasterPaddingAllowNegativeSpace();
```
mgmax, line 1225: see above

src/main/java/de/thomas_oster/liblasercut/drivers/GenericGcodeDriver.java
```diff
@@ -1207,8 +1288,14 @@ public void setProperty(String attribute, Object value) {
       this.setUploadMethod(value);
     } else if (SETTING_RASTER_PADDING.equals(attribute)) {
       this.setRasterPadding(Math.abs((Double)value));
+    } else if (SETTING_RASTER_PADDING_ALLOW_NEGATIVE_SPACE.equals(attribute)) {
+      this.setRasterPaddingAllowNegativeSpace((Boolean) value);
```
mgmax, line 1292: see above

src/main/java/de/thomas_oster/liblasercut/drivers/GenericGcodeDriver.java
```diff
@@ -494,45 +500,73 @@ protected void setSpeed(double speedInPercent) {
...
+    String formatPattern = "###.##";
+    DecimalFormat coordinateFormat = (DecimalFormat)NumberFormat.getNumberInstance(locale);
+    coordinateFormat.applyPattern(formatPattern);
+    coordinateFormat.setMaximumFractionDigits(getGCodeDigits());
+
     if (blankLaserDuringRapids)
     {
-      currentPower = 0.0;
-      sendLine("G0 X%f Y%f F%d S0", x, y, (int) (travel_speed));
+      currentPower = -1;
```
mgmax, line 534:
```suggestion
      currentPower = -1; // set to invalid value to force new S-value at next G1
```

src/main/java/de/thomas_oster/liblasercut/LaserCutter.java
```diff
@@ -315,6 +315,10 @@ public double getRasterPadding() {
       return 5;
     }
 
+    public boolean getRasterPaddingAllowNegativeSpace() {
+      return false;
+    }
+
```
mgmax, line 321: this should wait for a separate pull request

src/main/java/de/thomas_oster/liblasercut/LaserCutter.java
```diff
@@ -343,7 +347,17 @@ protected VectorPart convertRasterizableToVectorPart(RasterizableJobPart rp, Las
...
+      int leftLimitPx = 0;
+      if (this.getRasterPaddingAllowNegativeSpace())
+      {
+        leftLimitPx = (int) Util.mm2px(job.getTransformedOriginX() - this.getRasterPadding(), resolution);
+      }
+      else
+      {
+        leftLimitPx = (int) Util.mm2px(job.getTransformedOriginX(), resolution);
+      }
+
```
mgmax, line 360: this should wait for a separate pull request

src/main/java/de/thomas_oster/liblasercut/drivers/GenericGcodeDriver.java
```diff
@@ -1044,6 +1078,22 @@ public double getRasterPadding() {
...
+   * @return the value of rasterPaddingAllowNegativeSpace
+   */
+  @Override
+  public boolean getRasterPaddingAllowNegativeSpace() {
+    return rasterPaddingAllowNegativeSpace;
+  }
+
+  public void setRasterPaddingAllowNegativeSpace(boolean allowNegativeSpace) {
+    this.rasterPaddingAllowNegativeSpace = allowNegativeSpace;
+  }
```
mgmax, line 1096: see earlier

src/main/java/de/thomas_oster/liblasercut/drivers/GenericGcodeDriver.java
```diff
@@ -1084,7 +1156,10 @@ public void setApiKey(String apiKey)
     SETTING_FILE_EXPORT_PATH,
     SETTING_USE_BIDIRECTIONAL_RASTERING,
     SETTING_RASTER_PADDING,
-    SETTING_API_KEY
+    SETTING_RASTER_PADDING_ALLOW_NEGATIVE_SPACE,
```
mgmax, line 1159: see earlier

mgmax (changes requested):
we're almost there :-)

Did you also try if it resets to the correct default number of decimals if you save the setting with an old VisiCut version and then load it with the new one?

src/main/java/de/thomas_oster/liblasercut/drivers/GenericGcodeDriver.java
```diff
@@ -104,7 +107,10 @@
   protected static final String SETTING_SPINDLE_MAX = "S value for 100% laser power";
   protected static final String SETTING_UPLOAD_METHOD = "Upload method";
   protected static final String SETTING_RASTER_PADDING = "Extra padding at ends of raster scanlines (mm)";
+  protected static final String SETTING_RASTER_PADDING_ALLOW_NEGATIVE_SPACE = "Raster padding into negative coordinates";
```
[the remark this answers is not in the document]
McNugget6750: Agreed. I'll see if I can learn how to uncommit this and push this to a new branch.

McNugget6750:
> Did you also try if it resets to the correct default number of decimals if you save the setting with an old VisiCut version and then load it with the new one?

When I load a config that contains a valid laser cutter that does **not** have these values they will initialize to 0 digits. That's obviously undesirable. When you create a new laser cutter they will default to 6 decimals for G-codes and 0 decimals for S-codes. But I can change that to whatever you prefer. I prefer 0 as I have not seen any sample gCode file that's remotely compatible with GRBL or others that uses decimals on S-Codes.

How does the import work? Can you please point me to where I need to look to get defaults loaded when they are not part of the file? Is there such thing? I don't want to commit anything before the defaults load properly from an old settings file. Please help :)

I also reverted my fork back to before the negative space padding and will add my changes afterwards through a new branch.

McNugget6750:
Okey I give up looking. I spend about 90 minutes going through the code and finally I see this section in GenericGcodeDriver.java:1271 after reading all visicut.managers.

```
  /**
   * Adjust defaults after deserializing driver from an old version of XML file
   */
  @Override
  protected void setKeysMissingFromDeserialization()
  {
    System.out.println("Testing whether we ever reach setKeysMissingFromDeserialization:");
    System.out.println("this.spindleMax: " + this.spindleMax);
    System.out.println("this.gCodeDigits: " + this.gCodeDigits);
    System.out.println("this.sCodeDigits: " + this.sCodeDigits);
    // added field spindleMax, needs to be set to 1.0 by default
    // but xstream initializes it to 0.0 when it is missing from XML
    if (this.spindleMax <= 0.0) this.spindleMax = 1.0;
    if (this.gCodeDigits <= 0) this.gCodeDigits = 6;
    if (this.sCodeDigits <= 0) this.sCodeDigits = 0;
  }
```

But according to the debugger this code is never reached and "not part of a class", my printouts don't show up and the values are not being used. Where is this usually handled if not there? I tried debugging this as well as running this as normal and it won't show up with gCodeDigits set to 6. What am I missing?

t-oster:
Hi,
The simplest solution is to use Integer instead of int, because it will default to null not 0. Then you can check in the getter if it is null and set it to a sane default then. Obviously you would have to use the getter instead of the field directly in your code for that to work.

McNugget6750:
Thank you!
I did that, it works! I changed int to Integer and added a check it it's null. That does work with old settings files and I committed and pushed the changes to master. It should be good now.

mgmax (changes requested):
We're almost there. Looks very good. By the way, I found some bugs outside of your code which I will fix after it is merged.

src/main/java/de/thomas_oster/liblasercut/drivers/GenericGcodeDriver.java
```diff
@@ -494,45 +508,51 @@ protected void setSpeed(double speedInPercent) {
...
   protected void move(PrintStream out, double x, double y, double resolution) throws IOException {
     x = isFlipXaxis() ? getBedWidth() - Util.px2mm(x, resolution) : Util.px2mm(x, resolution);
     y = isFlipYaxis() ? getBedHeight() - Util.px2mm(y, resolution) : Util.px2mm(y, resolution);
     currentSpeed = getTravel_speed();
+
     if (blankLaserDuringRapids)
     {
-      currentPower = 0.0;
-      sendLine("G0 X%f Y%f F%d S0", x, y, (int) (travel_speed));
+      currentPower = -1; // set to invalid value to force new S-value at next G1
```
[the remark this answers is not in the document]
McNugget6750: Done

src/main/java/de/thomas_oster/liblasercut/drivers/GenericGcodeDriver.java
```diff
@@ -494,45 +508,51 @@ protected void setSpeed(double speedInPercent) {
...
-      sendLine("G0 Z%f", focus);
-      currentFocus = focus;
+
+    if (currentFocus != focus) {
+       String append = "";
+       if (blankLaserDuringRapids) {
+          append = " S0";
+          currentPower = -1; // set to invalid value to force new S-value at next G1
+       }
+       sendLine("G0 Z%s" + append, formatDouble(focus, getGCodeDigits()));
```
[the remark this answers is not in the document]
McNugget6750: Done

mgmax (approved):
Thanks a lot for your improvements in your first contribution to VisiCut! It would be great if you could continue these improvements.

PR #176, opened 2021-12-07
Negative overscan padding to allow tool head to move into negative X coordinates by 'extra padding mm'

McNugget6750:
Created a new branch for this feature to be able to open a new pull request.

Added feature to allow overscan padding into negative space. This defaults to "false" when created or when an old configuration file is loaded as boolean appear to initialize to 0/false.

Negative space padding allows the tool head to move into negative X values by the amount set for padding distance. Default is 5mm. Machine space home should therefore never be physical home when used unless physical home is not a hard end-stop. This is strictly machine specific and each user should decide for themselves if this feature is useful or not. However, most machines will have an offset defined when homing such that the head is a few mm away from physical endstops after $H (Home).

In my case, I move the toolhead around all the time and redefine where X0Y0 is and even invert the X axis for an easier, positive coordinate space workflow. So the physical limitation of not being allowed to move into negative X axis space is entirely arbitrary.


mgmax (changes requested):
Please rebase your branch to the current master. (Or, if that is too complicated, just recreate the change on top of the current master branch).

Optionally, please think about whether it makes sense to allow the same also into positive coordinates. The Epilog and LTT lasers (non-GCode) can overscan both left and right of the laser bed.

McNugget6750:
I'm looking into this again, now. I can merge to master no problem. I finally got git :)

However, positiv overscanning is less intuitive than negative overscanning which essentially worked out of the box by changing one line of code as submitted in this pull request. I thought I'd do the same for positive overscanning but it's not working as expected, yet. Any thoughts?

```
      if (this.getRasterPaddingAllowNegativeSpace())
      {
        leftLimitPx = (int) Util.mm2px(job.getTransformedOriginX() - this.getRasterPadding(), resolution);
        rightLimitPx = (int) Util.mm2px(job.getTransformedOriginX() + getBedWidth() **+ this.getRasterPadding()**, resolution);
      }
      else
      {
        leftLimitPx = (int) Util.mm2px(job.getTransformedOriginX(), resolution);
        rightLimitPx = (int) Util.mm2px(job.getTransformedOriginX() + getBedWidth(), resolution);
      }
```

McNugget6750:
Sorry, my bad. Maybe if I want to test the code, I also have to turn the feature on... It works now. I'll merge to master and push.

McNugget6750:
I closed this pull request as I implemented the requested changes and rebased to Master.

PR #179, opened 2021-12-16
Allow overscan padding outside machine space to allow tool head to move over the machine limits

McNugget6750:
Added feature to allow overscan padding outside machine space. This defaults to "false" when created or when an old configuration file is loaded as boolean appear to initialize to 0/false.

Negative space padding allows the tool head to move into negative X values by the amount set for padding distance. 
Positive space padding allows the tool head to move into positive X values beyond the max machine limits by the amount set for padding distance. Default is 5mm. Machine space home should therefore never be physical home when used unless physical home is not a hard end-stop. Same for positive max bed size. This is strictly machine specific and each user should decide for themselves if this feature is useful or not. However, most machines will have an offset defined when homing such that the head is a few mm away from physical endstops after $H (Home).

In my case, I move the toolhead around all the time and redefine where X0Y0 is and even invert the X axis for an easier, positive coordinate space workflow. So the physical limitation of not being allowed to move into negative X axis space is entirely arbitrary.


t-oster:
Easiest would be 'git fetch; git reset --mixed origin/master; git commit -a; git push -f <your remote here>;'

McNugget6750:
I ran the commands but nothing seems to have happened...

McNugget6750:
For the future, I'll work my way through this and learn how squashing works https://www.git-tower.com/learn/git/faq/git-squash.

mgmax:
My recommendation from personal experience:
- Do not use the `master` branch for PRs; use a separate branch for each thing you work on. 
- To clean up the history a PR you can use interactive rebase (or any other way to change the branch history) and then force-push. Then there's no need to close and reopen a new PR. It's a little bit of black magic, but you get used to it.

McNugget6750:
Did I misunderstand you in the other PR? https://github.com/t-oster/LibLaserCut/pull/176#pullrequestreview-828074762
I understood you asked me to merge into Master and then create a new PR...

mgmax:
Ah, sorry, so maybe my wording was confusing. The usual approach is that (e.g.) you a branch like `improve-whatever` or `fix-bug-123` to develop an improvement. Here, it's best if you don't put that into the `master` branch of your repository because you can't have multiple branches with the same name in one repository. At some point you clean up the history and submit a (draft) PR. If in the meanwhile the original repository (usually the `master` branch that your changes will later be merged into) has changed, then it is good to rebase and force-push. But still pushing that to your `improve-whatever` branch.

t-oster:
Hi,

I tried rebasing your changes to the current master of LibLaserCut, but there I can see you partially added your feature to the LaserCutter class and another part to the GenericGcodeDriver (see this commit https://github.com/t-oster/LibLaserCut/commit/6ee55605b005c82791796ef2cc92c65c68e82820 I will delete the branch after your code is merged).
I think it should be completely in the LaserCutter class because this class contains the raster-to-vector conversion.

Since there are only a few lines, please:
- Add our upstream repository to your remotes: `git remote add upstream git@github.com:t-oster/LibLaserCut.git`
- Create a new branch for your pull-request `git checkout -b feature-overscan-padding`
- Reset the branch (loosing all your changes!) to the latest master `git reset --hard upstream/master`
- Redo you changes in the LaserCutter.java
- Push them to your own repository in order to create a new pull request `git push origin feature-overscan-padding`

t-oster:
sorry for the inconvenience, but we have to keep the git history as clean as possible

mgmax:
I don't think it is right to put everything in the LaserCutter class. Many cutters, for example Epilog and LTT, derive from LaserCutter without using the raster-to-vector conversion. For these we don't want to have a user-visible config option.

In my opinion, the way the PR is currently implemented is correct: Put all the functionality in LaserCutter, but leave it disabled by default. Enable it in GenericGCodeDriver by overriding `getRasterPaddingAllowOutsideMachineSpace`.

t-oster:
Ok, that's a valid point. If you both agree I will then just commit my squashed Version in behalf of @McNugget6750 .

mgmax:
looks good. Some documentation comment would be helpful, but I can also add that later after merging. For example:

`getRasterPaddingAllowOutsideMachineSpace()`
> When engraving, is it allowed that the padding (overscan) goes beyond the laser bed width? This only affects convertRasterizableToVectorPart(). It has no effect on laser drivers that use a different implementation of raster-engraving, e.g., the EpilogCutter driver.

t-oster:
OK @McNugget6750 thanks for your contribution. Please test if everything is working as expected in the latest VisiCut release (should be online in a few minutes).

McNugget6750:
hey guys, thanks for the extra effort! I will be out of office for the next two weeks, so I will not be able to test it until then. Since the changes were quite minor and only affected a few lines of code, I don’t expect any issues. I will look at the newly merged code but I will not be able to test it until I’m back.

And yes, I agree, it should be enabled and disabled from inside the generic gcode driver.

PR #186, opened 2023-01-12
Fix ruida rasterizable

### LINKED ISSUE #187: Engrave with Ruida driver goes wild
Seen with 
VisiCut		 Version: 1.9-169-g99244adc
LibLaserCut	 Version: visicut1.7
on Ubuntu 22.04 
using Model: Ruida	 Driver:de.thomas_oster.liblasercut.drivers.Ruida

I am loading this PNG image into visicut, scale it to 6 cm x 5.6 cm.
Engrave profile, with default 500dpi, floyd steinberg, Grayscale shift 0, not inverted colors.
Use bidirectional rastering [x]
min power 20, max power 20, speed 100.

schickard_portrait

I'd expect engraving line by line with even speed and some reasonable overshoot on both sides.
I see jerky movement, decelerating and accelerating in mid flight, no overshoot at the sides, with some occasional diagonal and vertical movements. 

https://user-images.githubusercontent.com/1108546/212476800-fdbe92cb-5358-44f9-8c87-d73b8b98f142.mp4

The attached zip contains a .rd file saved from such an attempt.
[schickard.zip](https://github.com/t-oster/LibLaserCut/files/10417906/schickard.zip)

kkaempf:
Fixes #187

### DIFF
--- a/src/main/java/de/thomas_oster/liblasercut/FloatMinMaxPowerSpeedFrequencyProperty.java
+++ b/src/main/java/de/thomas_oster/liblasercut/FloatMinMaxPowerSpeedFrequencyProperty.java
@@ -31,7 +31,6 @@ import java.util.LinkedList;
 public class FloatMinMaxPowerSpeedFrequencyProperty extends FloatPowerSpeedFrequencyProperty {
 
   private float min_power = 10.0f;
-  private float speed = 500.0f;
   private static final String MIN_POWER = "min power";
 
   public FloatMinMaxPowerSpeedFrequencyProperty()
--- a/src/main/java/de/thomas_oster/liblasercut/FloatMinMaxPowerSpeedFrequencyProperty.java
+++ b/src/main/java/de/thomas_oster/liblasercut/FloatMinMaxPowerSpeedFrequencyProperty.java
@@ -33,7 +33,6 @@ public class FloatMinMaxPowerSpeedFrequencyProperty extends FloatPowerSpeedFrequ
   private float min_power = 10.0f;
   private float speed = 500.0f;
   private static final String MIN_POWER = "min power";
-  private static final String SPEED = "speed";
 
   public FloatMinMaxPowerSpeedFrequencyProperty()
   {
@@ -90,31 +89,9 @@ public class FloatMinMaxPowerSpeedFrequencyProperty extends FloatPowerSpeedFrequ
     if (MIN_POWER.equals(name)) {
       return (Float) this.getMinPower();
     }
-    else if (SPEED.equals(name)) {
-      return (Float) this.getSpeed();
-    }
     else {
       return super.getProperty(name);
     }
-  }
-
-  /**
-   * Must override since parent class only allows 0 < speed < 100
-   * 
-   */
-
-  @Override
-  public void setSpeed(float speed)
-  {
-    speed = speed < 0 ? 0 : speed;
-    speed = speed > 1000 ? 1000 : speed;
-    this.speed = speed;
-  }
-
-  @Override
-  public float getSpeed()
-  {
-    return speed;
   }
 
   @Override
@@ -144,9 +121,6 @@ public class FloatMinMaxPowerSpeedFrequencyProperty extends FloatPowerSpeedFrequ
   {
     if (MIN_POWER.equals(name)) {
       this.setMinPower((Float) value);
-    }
-    else if (SPEED.equals(name)) {
-      this.setSpeed((Float) value);
     }
     else {
       super.setProperty(name, value);
--- a/src/main/java/de/thomas_oster/liblasercut/FloatMinMaxPowerSpeedFrequencyProperty.java
+++ b/src/main/java/de/thomas_oster/liblasercut/FloatMinMaxPowerSpeedFrequencyProperty.java
@@ -1,7 +1,7 @@
 /**
  * This file is part of LibLaserCut.
  * Copyright (C) 2011 - 2014 Thomas Oster <mail@thomas-oster.de>
- * Copyright (C) 2018 - 2020 Klaus Kämpf <kkaempf@suse.de>
+ * Copyright (C) 2018 - 2023 Klaus Kämpf <kkaempf@gmail.com>
  *
  * LibLaserCut is free software: you can redistribute it and/or modify
  * it under the terms of the GNU Lesser General Public License as published by
--- a/src/main/java/de/thomas_oster/liblasercut/FloatMinMaxPowerSpeedFrequencyProperty.java
+++ b/src/main/java/de/thomas_oster/liblasercut/FloatMinMaxPowerSpeedFrequencyProperty.java
@@ -31,7 +31,9 @@ import java.util.LinkedList;
 public class FloatMinMaxPowerSpeedFrequencyProperty extends FloatPowerSpeedFrequencyProperty {
 
   private float min_power = 10.0f;
+  private float speed = 500.0f;
   private static final String MIN_POWER = "min power";
+  private static final String SPEED = "speed";
 
   public FloatMinMaxPowerSpeedFrequencyProperty()
   {
@@ -88,9 +90,31 @@ public class FloatMinMaxPowerSpeedFrequencyProperty extends FloatPowerSpeedFrequ
     if (MIN_POWER.equals(name)) {
       return (Float) this.getMinPower();
     }
+    else if (SPEED.equals(name)) {
+      return (Float) this.getSpeed();
+    }
     else {
       return super.getProperty(name);
     }
+  }
+
+  /**
+   * Must override since parent class only allows 0 < speed < 100
+   * 
+   */
+
+  @Override
+  public void setSpeed(float speed)
+  {
+    speed = speed < 0 ? 0 : speed;
+    speed = speed > 1000 ? 1000 : speed;
+    this.speed = speed;
+  }
+
+  @Override
+  public float getSpeed()
+  {
+    return speed;
   }
 
   @Override
@@ -120,6 +144,9 @@ public class FloatMinMaxPowerSpeedFrequencyProperty extends FloatPowerSpeedFrequ
   {
     if (MIN_POWER.equals(name)) {
       this.setMinPower((Float) value);
+    }
+    else if (SPEED.equals(name)) {
+      this.setSpeed((Float) value);
     }
     else {
       super.setProperty(name, value);
--- a/src/main/java/de/thomas_oster/liblasercut/drivers/Ruida.java
+++ b/src/main/java/de/thomas_oster/liblasercut/drivers/Ruida.java
@@ -92,8 +92,8 @@ public class Ruida extends LaserCutter
   protected static final String SETTING_BED_WIDTH = "Bed width (mm)";
   protected static final String SETTING_BED_HEIGHT = "Bed height (mm)";
   protected static final String SETTING_USE_BIDIRECTIONAL_RASTERING = "Use bidirectional rastering";
-  protected static final String SETTING_RASTER_PADDING = "Padding per raster line (mm)";
-  protected static final String SETTING_RASTER_OUTSIDE = "Allow padding outside of specs";
+  protected static final String SETTING_RASTER_PADDING = "Extra padding at ends of raster scanlines (mm)";
+  protected static final String SETTING_RASTER_OUTSIDE = "Allow raster padding outside machine limits (negative and positive)";
   protected static final Locale FORMAT_LOCALE = Locale.US;
 
   protected static final String[] uploadMethodList = {UPLOAD_METHOD_FILE, UPLOAD_METHOD_IP, UPLOAD_METHOD_SERIAL};
--- a/src/main/java/de/thomas_oster/liblasercut/drivers/Ruida.java
+++ b/src/main/java/de/thomas_oster/liblasercut/drivers/Ruida.java
@@ -186,11 +186,11 @@ public class Ruida extends LaserCutter
    * 'runway' for laser to get up to speed when rastering (in mm)
    *
    */
-  private Double rasterPadding = 20;
+  private Double rasterPadding = 20.;
 
   @Override
-  public Double getRasterPadding() {
-    if (rasterPadding == null) rasterPadding = 20;
+  public double getRasterPadding() {
+    if (rasterPadding == null) rasterPadding = 20.;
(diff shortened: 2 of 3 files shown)


jnweiger:
Here you are: https://github.com/t-oster/LibLaserCut/issues/187

src/main/java/de/thomas_oster/liblasercut/FloatMinMaxPowerSpeedFrequencyProperty.java
```diff
@@ -88,11 +90,33 @@ public Object getProperty(String name)
...
+  /**
+   * Must override since parent class only allows 0 < speed < 100
+   * 
+   */
+
+  @Override
+  public void setSpeed(float speed)
+  {
+    speed = speed < 0 ? 0 : speed;
+    speed = speed > 1000 ? 1000 : speed;
```
mgmax, line 110: All speed settings for all other lasercutters are in percent, maximum 100. So this one should be the same for consistency. If we support absolute speeds or other units (mm/s, inch per whatever) one day, then this should be one consistent interface for all/most lasercutters, see https://github.com/t-oster/VisiCut/issues/381

src/main/java/de/thomas_oster/liblasercut/drivers/Ruida.java
```diff
@@ -180,6 +182,36 @@ public boolean canEstimateJobDuration()
     return false;
   }
 
+  /**
+   * 'runway' for laser to get up to speed when rastering (in mm)
+   *
+   */
+  private double rasterPadding = 20;
```
mgmax, line 189: Please double-check if this is initialized correctly when loading old settings that do not yet have this property. As far as I remember, XStream sets all missing fields to zero/false/null, ignoring the default value. You may need to implement it similar to here (`Double` instead of `double` datatype, then a missing initialization will be `null` and can be distinguished from `0`). https://github.com/t-oster/LibLaserCut/blob/5215383458060acb7404d01356a2a8af0137f52c/src/main/java/de/thomas_oster/liblasercut/drivers/GenericGcodeDriver.java#L1098-L1104

src/main/java/de/thomas_oster/liblasercut/drivers/Ruida.java
```diff
@@ -92,6 +92,8 @@ public class Ruida extends LaserCutter
   protected static final String SETTING_BED_WIDTH = "Bed width (mm)";
   protected static final String SETTING_BED_HEIGHT = "Bed height (mm)";
   protected static final String SETTING_USE_BIDIRECTIONAL_RASTERING = "Use bidirectional rastering";
+  protected static final String SETTING_RASTER_PADDING = "Padding per raster line (mm)";
+  protected static final String SETTING_RASTER_OUTSIDE = "Allow padding outside of specs";
```
mgmax, line 96: should be the same texts as in https://github.com/t-oster/LibLaserCut/blob/5215383458060acb7404d01356a2a8af0137f52c/src/main/java/de/thomas_oster/liblasercut/drivers/GenericGcodeDriver.java#L109-L110 unless there is a very good reason. If the other text is not good, feel free to change both.

mgmax (commented):
Nice to hear from you! I have some feedback, maybe it helps, maybe not ;-)

jnweiger:
Ah, historical ballast. Sorry. My original driver (as located in the private fork) used absolute values mm/sec.

I understand the consistency argument within visicut. But the same argument can be applied to ruida settings across visicut, lightburn, and rdworks. Both, lightburn and rdworks use mm/sec, so making it percent in visicut makes it inconsistent there.

mgmax:
> Both, lightburn and rdworks use mm/sec, so making it percent in visicut makes it inconsistent there.

Of course the UI should offer both in the future.  Until then, a local hack in one single driver will just cause confusion.
Currently, there is no proper API for absolute values (get units to show in the UI, conversion between abs. and percent).

kkaempf:
https://github.com/t-oster/LibLaserCut/pull/186/commits/01b069ebf5d72fc3663f1b431533c30085bc7938 🤷🏻‍♂️ 😆

src/main/java/de/thomas_oster/liblasercut/drivers/Ruida.java
kkaempf: ✅

kkaempf:
Tests were successful, marking "ready for review". Thanks ! 🙏🏻

src/main/java/de/thomas_oster/liblasercut/drivers/Ruida.java
mgmax: As far as I understand, the type should be `Double` not `double`. `Double` can be `null` (like null-pointer), `double` is just a value.

Please check that
- old settings that don't have the value are not loaded as 0 but as 20 (due to the `null` trick).
- in new settings, you can set the value to 0, save it, restart VisiCut, and it still remains at 0.

I'd be surprised if the current code actually works in all these cases.

mgmax (changes requested):

src/main/java/de/thomas_oster/liblasercut/drivers/Ruida.java
kkaempf: `double` -> `Double` ✅

jnweiger: This uppercase D in Double causes compile errors:

[ERROR] Failed to execute goal org.apache.maven.plugins:maven-compiler-plugin:3.1:compile (default-compile) on project liblasercut: Compilation failure: Compilation failure: 
[ERROR] /home/testy/src/github/kkaempf/LibLaserCut/src/main/java/de/thomas_oster/liblasercut/drivers/Ruida.java:[189,34] incompatible types: int cannot be converted to java.lang.Double
[ERROR] /home/testy/src/github/kkaempf/LibLaserCut/src/main/java/de/thomas_oster/liblasercut/drivers/Ruida.java:[192,17] getRasterPadding() in de.thomas_oster.liblasercut.drivers.Ruida cannot override getRasterPadding() in de.thomas_oster.liblasercut.LaserCutter
[ERROR]   return type java.lang.Double is not compatible with double
[ERROR] /home/testy/src/github/kkaempf/LibLaserCut/src/main/java/de/thomas_oster/liblasercut/drivers/Ruida.java:[191,3] method does not override or implement a method from a supertype
[ERROR] /home/testy/src/github/kkaempf/LibLaserCut/src/main/java/de/thomas_oster/liblasercut/drivers/Ruida.java:[193,48] incompatible types: int cannot be converted to java.lang.Double
[ERROR] -> [Help 1]

mgmax (approved):
looks good. I'll merge soon if there are no further comments.

src/main/java/de/thomas_oster/liblasercut/drivers/Ruida.java
mgmax: @kkaempf has this been fixed?
It should be enough to change the type the property but leave the functions unchanged.

mgmax (changes requested):
Please clarify the comment by @jnweiger .
Did you test two the mentioned cases regarding loading of settings?

jnweiger:
I can compile with this additional patch:

```
--- a/src/main/java/de/thomas_oster/liblasercut/drivers/Ruida.java
+++ b/src/main/java/de/thomas_oster/liblasercut/drivers/Ruida.java
@@ -186,11 +186,11 @@ public class Ruida extends LaserCutter
    * 'runway' for laser to get up to speed when rastering (in mm)
    *
    */
-  private Double rasterPadding = 20;
+  private Double rasterPadding = 20.;
 
   @Override
-  public Double getRasterPadding() {
-    if (rasterPadding == null) rasterPadding = 20;
+  public double getRasterPadding() {
+    if (rasterPadding == null) rasterPadding = 20.;
     return rasterPadding;
   }
```

src/main/java/de/thomas_oster/liblasercut/drivers/Ruida.java
kkaempf: Just merged @jnweiger's PR kkaempf/LibLaserCut#3

src/main/java/de/thomas_oster/liblasercut/drivers/Ruida.java
```diff
@@ -941,6 +982,10 @@ public void setProperty(String attribute, Object value) {
       this.setBedWidth((Double)value);
     } else if (SETTING_USE_BIDIRECTIONAL_RASTERING.equals(attribute)) {
       this.setUseBidirectionalRastering((Boolean) value);
+    } else if (SETTING_RASTER_PADDING.equals(attribute)) {
+      this.setRasterPadding((Double) value);
+    } else if (SETTING_RASTER_OUTSIDE.equals(attribute)) {
```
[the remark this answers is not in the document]
kkaempf: Using upstream standard

mgmax:
Code looks good.
Did you test the two cases I asked about earlier?

kkaempf:
> Did you test the two cases I asked about earlier?

Sorry no. And I wouldn't even know how.

Maybe @jnweiger can answer this ? 🤔

jnweiger:
I've done the following test:

* Install deb package visicut-1.9-171-ga45332c3
* start `/usr/bin/visicut`
* Create a 'Nova 35 fablabbg' device with the Ruida driver.
* Load an example svg, select 'cut all'
* Create some settings:
  - 3mm cut: minPower=33% maxPower=80% speed=50% freq=500
  - 4mm cut: minPower=44% maxPower=80% speed=40% freq=500
  - 5mm cut: minPower=55% maxPower=80% speed=30% freq=500
* Send to file with the 3mm setting.

Then compile github/t-oster/Visicut master with this github/kkaempf/LibLasercut fix-ruida-rasterizable 
* start `java -Xms256m -Xmx1024m -jar target/visicut-*-full.jar`
* Load an example svg,  select 'cut all'
* Inspect the settings:
    - 3mm cut: minPower=33% maxPower=80% speed=0% freq=500
    - 4mm cut: minPower=44% maxPower=80% speed=0% freq=500
    - 5mm cut: minPower=55% maxPower=80% speed=0% freq=500

The speeds all show 0.
Retesting with /usr/bin/visicut shows that the speeds were correctly saved.

jnweiger:
The second test works fine:

* with target/visicut-1.9-SNAPSHOT-full.jar change the minpower of the 5mm setting to 0.
* exit visicut, restart.
* the setting show up as 0 as expected.

jnweiger:
@mgmax what needs to be done here to get it merged?
I'd be much more comfortable to have a working visicut first, and then fix the details about using the correct 'double' datatype later.
Currently this PR blocks us completely from using visicut at the lab.

mgmax:
The only thing I am strict about here is that old settings (lasercutter device settings, material power/speed, ...) created using an old *mainline* VisiCut version still work in the new version. This includes that default values for variables that did not exist before this PR are initialized correctly. I don't care about compatibility with custom forks.

The XStream serializer we use has a bug/misfeature that if you define `class myClass { double foo = 123; }` and load an XML file that does not contain `foo` (e.g. from an older Visicut version) then you will get `foo=0`. Therefore the `Double` workaround is needed.

I agree that the sonatype check is only a warning and not relevant here.

src/main/java/de/thomas_oster/liblasercut/FloatMinMaxPowerSpeedFrequencyProperty.java
```diff
@@ -31,6 +31,7 @@
 public class FloatMinMaxPowerSpeedFrequencyProperty extends FloatPowerSpeedFrequencyProperty {
 
   private float min_power = 10.0f;
+  private float speed = 500.0f;
```
mgmax, line 34: Loading the speed from old settings does not work correctly for me. I guess this line should be removed.

mgmax (changes requested):

src/main/java/de/thomas_oster/liblasercut/FloatMinMaxPowerSpeedFrequencyProperty.java
mgmax: According to my first quick check, everything else (load and save device settings) seems to work.

The default power value 0 is correct and same as for the Epilog driver. (In the above discussion, "0 vs 20" referred to raster padding, not power.)

Regarding usability, you could possibly improve the editing of "min power" (in some future version, not in this PR). Currently, the default initial value is unlogical (min power = 10 > power = 0). When editing min power to e.g. 11 it is reset to 0, which technically makes sense but may be confusing to the user who wants to enter "min power" first and then "power" second. You could remove the limiting of "min power" from the Property class and instead just limit it in the driver (silently, or maybe with a warning message).

jnweiger: @kkaempf is it acceptable what @mgmax suggests?
I'd like to get this merged. This blocks us from using visicut with thunderlaser.
(my private build always comes up with an empty settings panel for unknown reasons)

mgmax:
I see that the driver converts speeds  0-100%  to 0-1000mm/s. Your driver has settings for "Max vector cutting speed (mm/s)" and "Max vector move speed (mm/s)" which currently have no effect. Do all Ruida machines have a limit of 100% = 1000 mm/s? 

If yes, these settings should probably be removed.
To do so, remove everything related to `MaxVector...` and `MAX_VECTOR_...` except for the one line where the class attribute is declared. To avoid that importing old settings fails, do not remove the attribute but mark it as `@Deprecated transient` , as in
https://github.com/t-oster/LibLaserCut/blob/bc3f41385535869d6c47f158ddb7fdb48a2997dd/src/main/java/de/thomas_oster/liblasercut/drivers/LaserToolsTechnicsCutter.java#L238-L244

If no, then the conversion should depend on the configured speed. `speed_mmps = speed_percent * max_speed / 100`

kkaempf:
Hmm, this issue started with "fix Ruida rasterizable" and has meanwhile grown to a generic (and helpful ! Don't get me wrong here 😉 ) Ruida driver improvement one.

I'd rather see this splitted into multiple issues and fix problems in separated PRs.

mgmax:
Good suggestion!
The leftover issue was detached to https://github.com/t-oster/LibLaserCut/issues/190 .
Furthermore, you could look into the (GUI) usability of "min power" as discussed before.

PR #191, opened 2023-08-31
Calibrate Epilog Helix constants for `estimateJobDuration`

sethtroisi:
I took 22 measurements with my Epilog Helix Mini doing cuts at several angles and moving in several angles and got very similar speeds. These are roughly 25% slower than the existing numbers.

I added some tests for both EpilogHelix and EpilogZing.

I also ran and tested this change in VisiCut (though my java setup in intellij is incomplete and failed to loading some of the XML but that's a different issue)

Screenshot from 2023-08-31 15-13-31

### DIFF
--- a/src/main/java/de/thomas_oster/liblasercut/drivers/EpilogCutter.java
+++ b/src/main/java/de/thomas_oster/liblasercut/drivers/EpilogCutter.java
@@ -964,22 +964,6 @@ abstract class EpilogCutter extends LaserCutter
     return true;
   }
 
-  @Override
-  public int estimateJobDuration(LaserJob job)
-  {
-    // It's not entirely clear how these values were determined. They are probably calibrated for the Epilog Zing.
-    double PX2MM_500DPI = Util.px2mm(1, 500);
-    double VECTOR_MOVESPEED_X = PX2MM_500DPI * 20000d / 4.5;
-    double VECTOR_MOVESPEED_Y = PX2MM_500DPI * 10000d / 2.5;
-    double VECTOR_LINESPEED = PX2MM_500DPI * 20000d / 36.8;
-    double RASTER_LINEOFFSET = 0.08d;
-    double RASTER_LINESPEED = PX2MM_500DPI * 100000d / ((268d / 50) - RASTER_LINEOFFSET);
-    //TODO: The Raster3d values are not tested yet, theyre just copies
-    double RASTER3D_LINEOFFSET = RASTER_LINEOFFSET;
-    double RASTER3D_LINESPEED = RASTER_LINESPEED;
-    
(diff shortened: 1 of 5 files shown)


src/main/java/de/thomas_oster/liblasercut/drivers/EpilogHelix.java
```diff
@@ -74,76 +83,18 @@ public EpilogHelix clone()
...
-  
-  @Override
-  public boolean isHideSoftwareFocus()
-  {
-    return super.isHideSoftwareFocus();
-  }
+  public int estimateJobDuration(LaserJob job)
+  {
+    // TODO: Calibrate these RASTER settings.
+    double PX2MM_500DPI = Util.px2mm(1, 500);
```
sethtroisi, line 90: If this change is accepted I'll spend the time to try to calibrate these settings which produce very wrong estimates IMO.

t-oster:
Hi, thanks for your work. Looks good. However it seems as is you are doing unnecessary formatting and moving methods around, so that the changes are not as clearly visible in the diff as they could (see https://github.com/t-oster/LibLaserCut/pull/191/files).
You also do not need to remove the method from EpilogCutter in order to override it in the specific EpilogHelix class.
Could you apply your changes to a clean checkout without changing anything else?

t-oster:
You also removed many methods, which are not necessary from a java standpoint of view, but the comment suggests they are needed for the XML encoder, so not sure if this breaks anything.

sethtroisi:
I reverted the unnecessary method removal and will pursue that in a separate PR. I've tried to distill this to just the vital pieces.

src/main/java/de/thomas_oster/liblasercut/drivers/EpilogCutter.java
```diff
@@ -964,22 +964,6 @@ public boolean canEstimateJobDuration()
...
-    double VECTOR_MOVESPEED_Y = PX2MM_500DPI * 10000d / 2.5;
-    double VECTOR_LINESPEED = PX2MM_500DPI * 20000d / 36.8;
-    double RASTER_LINEOFFSET = 0.08d;
-    double RASTER_LINESPEED = PX2MM_500DPI * 100000d / ((268d / 50) - RASTER_LINEOFFSET);
-    //TODO: The Raster3d values are not tested yet, theyre just copies
-    double RASTER3D_LINEOFFSET = RASTER_LINEOFFSET;
-    double RASTER3D_LINESPEED = RASTER_LINESPEED;
-    
-    return estimateJobDuration(job, VECTOR_MOVESPEED_X, VECTOR_MOVESPEED_Y, VECTOR_LINESPEED, RASTER_LINEOFFSET, RASTER_LINESPEED, RASTER3D_LINEOFFSET, RASTER3D_LINESPEED);
-  }
```
sethtroisi, line 982: I felt it was better to move the printer specific parameters to `EpilogZing` and `EpilogHelix` but if any external user is subclassing `EpilogCutter` (and doesn't override) they would now get `RuntimeException("Method not implemented");` .

Issue #194, opened 2023-09-25
Ruida driver does not change between Mark and Cut


jnweiger:
@kkaempf
Configure green lines for Mark (weak laser power and fast move, to scratch only the surface)
Configure red lines for Cut (strong and slow)

Load an svg that uses both, e.g. the attached head.svg 
![head](https://github.com/t-oster/LibLaserCut/assets/1108546/5e32f9b8-23d7-4182-915e-fa2da6e18ade)

When Mark is first in the list, then everything is marked.
When Cut is first in the list, then everything is cut.

Expected behaviour: when done with one setting, we correctly switch to the other.

jnweiger:
The initialization of the first layer "green, mark" is
```
Laser_1_Min_Pow_C6_31 Layer:0 10%                       c6 31 00 0c 66 
Laser_1_Max_Pow_C6_32 Layer:0 35%                       c6 32 00 2c 66
Layer_Speed Layer:0 200.0mm/s                           c9 04 00 00 00 0c 1a 40
Laser_Freq Laser1 00 0.5kHz                             c6 60 00 00 00 00 00 03 74
Layer_Color Layer:0 R 39%, G 0%, B 0%                   ca 06 00 00 00 00 00 64 
Layer_CA_41 Layer:0 00                                  ca 41 00 00 
Flags_CA_01 00                                          ca 01 00 
CA 02 Prio 0                                            ca 02 00
Blow_on                                                 ca 01 13 
Speed_C9 0.2mm/s                                        c9 02 00 00 00 01 48 
Laser_1_Min_Pow_C6_01 10%                               c6 01 0c 66 
Laser_1_Max_Pow_C6_02 35%                               c6 02 2c 66
CA_03 0f                                                ca 03 0f
CA 10 00                     
```

The initialization of the second layer "red, cut" is
```
Laser_1_Min_Pow_C6_01 50%                               c6 01 3f 7f
Laser_1_Max_Pow_C6_02 80%                               c6 02 66 33
Speed_C9 14.0mm/s                                       c9 02 00 00 00 6d 30
```

jnweiger:
With "my" old visicut 1.8, th initialization looked like this:
First layer: 
```
Layer_Speed Layer:0 200.0mm/s                           c9 04 00 00 00 0c 1a 40
Laser_1_Min_Pow_C6_31 Layer:0 9%                        c6 31 00 0b 42
[...]

Issue #190, opened 2023-06-28
Ruida driver speed conversion wrong?


mgmax:
The ruida driver converts speeds  0-100%  to 0-1000mm/s. This value is hardcoded, even though the driver has settings for "Max vector cutting speed (mm/s)" and "Max vector move speed (mm/s)". These settings currently have no effect.

Do all Ruida machines have a limit of 100% = 1000 mm/s? 

If yes, these settings should probably be removed.
To do so, remove everything related to `MaxVector...` and `MAX_VECTOR_...` except for the one line where the class attribute is declared. To avoid that importing old settings fails, do not remove the attribute but mark it as `@Deprecated transient` , as in
https://github.com/t-oster/LibLaserCut/blob/bc3f41385535869d6c47f158ddb7fdb48a2997dd/src/main/java/de/thomas_oster/liblasercut/drivers/LaserToolsTechnicsCutter.java#L238-L244

If no, then the conversion should depend on the configured speed. `speed_mmps = speed_percent * max_speed / 100`

_Originally posted by @mgmax in https://github.com/t-oster/LibLaserCut/issues/186#issuecomment-1605588288_

fblaese:
The speed hardware limit is configurable via RDWorks. Instead of removing this setting, it should be utilized for the speed conversion.
I will send a Pull Request for this (and a few other improvements) soon.

On a side note: I think that setting the speed as a percentage of the machine limits is quite counterintuitive. I would prefer an absolute speed setting in mm/s, which is automatically confined to the machine limits.

Issue #189, opened 2023-03-29
Cut Start with Ruida starts from the origin, instead of first vertex


jnweiger:
Seen with Visicut 1.9.0 compiled with liblasercut from https://github.com/t-oster/LibLaserCut/pull/186

The attached SVG should produce a regular 6 pointed star. 
The preview in visicut shows the star correctly, 
but when sent to the laser, the first line starts at the top left origin, instead of at the first vertex.
It is shown incorrect in the machine display, and it also cuts incorrectly.
<img src="https://user-images.githubusercontent.com/1108546/228678658-04794f96-1ddc-41e5-a347-797f56b4746b.jpg" width="300" />&nbsp;<img src="https://user-images.githubusercontent.com/1108546/228678684-115939e8-249c-4576-893f-b96ff5582284.jpg" width="300" />

Assign to @kkaempf 

![Zeichnung-1-17](https://user-images.githubusercontent.com/1108546/228960738-9be92951-e47d-4121-a821-3228f9482c9d.svg)

t-oster:
Hi,
does this also occur if you compile without #186. Is the problem only in the ruida driver, or can you confirm it in other drivers?

jnweiger:
ah, good question!
I should have a chance to verify that with the zing driver soonish...

kkaempf:
Hmm, disassembling the generated `.rd` file from `Ruida.java`:

```
...
Blow_on							ca 01 13 
Speed_C9 0.1mm/s					c9 02 00 00 00 00 64 
Laser_1_Min_Pow_C6_01 10%				c6 01 0c 66 
Laser_1_Max_Pow_C6_02 0%				c6 02 00 00 
CA_03 0f						ca 03 0f 
CA 10 00						ca 10 00 
Mov_Abs 49.222mm 46.213mm				88 00 00 03 00 46 00 00 02 69 05 
Cut_Abs 31.351mm 40.147mm				a8 00 00 01 74 77 00 00 02 39 53 
Cut_Abs 22.541mm 56.837mm				a8 00 00 01 30 0d 00 00 03 3c 05 
```

comparing this with https://github.com/t-oster/LibLaserCut/blob/bc3f41385535869d6c47f158ddb7fdb48a2997dd/src/main/java/de/thomas_oster/liblasercut/drivers/Ruida.java#L585-L595
I'd say that the initial MOVE is coming from LibLaserCut.

t-oster:
[...]

PR #197, opened 2023-12-03
Ruida: various driver bug fixes and improvments

fblaese:
Fix a few bugs in the Ruida driver:
- Fix 0xC902 speed conversion
- Calculate speed percentage from maximum cutting speed (#190)
- Remove unnecessary array stop command
- Fix layer property code generation (#194)
- Force movement if last position is unknown (#189)
- Add missing reset of instance variables

### DIFF
--- a/src/main/java/de/thomas_oster/liblasercut/drivers/Ruida.java
+++ b/src/main/java/de/thomas_oster/liblasercut/drivers/Ruida.java
@@ -1066,7 +1066,7 @@ class ByteStream
    * append relative value
    */
   public ByteStream relative(double d, boolean signed) throws IOException {
-    int val = (int)Math.floor(d);
+    int val = (int)Math.round(d);
 //    System.out.println("rel" + ((signed)?"Signed":"Unsigned") + "ValueToByteArray(" + d + " -> " + val + ")");
     if (signed) {
       if (val > 8191) {
--- a/src/main/java/de/thomas_oster/liblasercut/drivers/Ruida.java
+++ b/src/main/java/de/thomas_oster/liblasercut/drivers/Ruida.java
@@ -92,9 +92,6 @@ public class Ruida extends LaserCutter
   protected static final String SETTING_MAX_POWER = "Max laser power (%)";
   protected static final String SETTING_BED_WIDTH = "Bed width (mm)";
   protected static final String SETTING_BED_HEIGHT = "Bed height (mm)";
-  protected static final String SETTING_USE_BIDIRECTIONAL_RASTERING = "Use bidirectional rastering";
-  protected static final String SETTING_RASTER_PADDING = "Extra padding at ends of raster scanlines (mm)";
-  protected static final String SETTING_RASTER_OUTSIDE = "Allow raster padding outside machine limits (negative and positive)";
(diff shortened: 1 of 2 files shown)


mgmax:
Is it intentional that you removed the raster padding settings introduced in https://github.com/t-oster/LibLaserCut/pull/186 https://github.com/t-oster/LibLaserCut/issues/187 ?

I'm a little confused because the get/set methods and comments are still there but they are used nowhere.

When "removing" variables, please make them `@Deprecated transient`. Please also add a comment like "  // This variable is only for compatibility with imported old settings".

fblaese:
Yes, this is indeed intentional. The controller automatically generates appropriately sized padding to reach the configured raster speed when using the sweep movement modes, therefore no manually generated raster padding is required.
The sweep modes probably have not been discovered by reverse-engineering before.

The setters for `rasterPadding, `allowOutsidePadding` and `useBidirectionalRastering` are already removed by this pull request. The variables are kept and annotated with `@Deprecated` to keep configuration files compatible.
The getters for `rasterPadding` and `allowOutsidePadding` are kept to override the generic getters of the parent class.

The getter for `useBidirectionalRastering` is kept because bidirectional rastering should still be possible with my changes. Though, the controllers behavior in the unidirectional sweep modes has not been tested thoroughly yet, so unidirectional rastering is disabled. I might test and reenable unidirectional rastering in the future.

I will add the `transient` modifier to the deprecated variables.

fblaese:
I have added the `transient` modifier to the deprecated variables.

jnweiger:
@mgmax are we ready to merge here?

jnweiger:
@fblaese The scaling factor for speed seems to be 11.6 now. Speed 1.3 in your visicut shows up as 15 mm/sec on the machine. 1.8 as 21.
We used to have exactly 10 as scaling factor, I liked that better :-) but nevermind, lets merge now and discuss that later!

mgmax:
I'm slightly unhappy that this change breaks old laser settings by making them 10% faster (which typically means that the material will be almost cut through instead of fully cut through). But I guess that's the price we have to pay for faster development :-)

@fblaese is it okay for you to merge? Then I will do so.

fblaese:
@jnweiger Previously the scaling factor from "speed percentage" to "actual movement speed in mm/s" was fixed at 10 - derived from a maximum movement speed (100%) of 1000 mm/s. With this pull request, the speed percentage is derived from the maximum movement speed in the laser settings. If the maximum movement speed inside the laser settings is changed, the scaling factor changes as well (see #190 and cdb9efc7f930e27c4067f5b2a5ada5aae5d60f32).

The default maximum travel/cutting speed for the Ruida driver is 1000 mm/s though, so without any changes to the default settings, the factor should be identical to previous versions. (You probably have changed the maximum speed settings?)

The pull request should be ready to merge.
However, I have noticed that the laser driver has two seperate max speeds. One for movement (laser off) and one for cuts. My change currently do not respect that, but always calculates the speed percentage scaling factor from the maximum cutting speed. I can update this PR in the next few days to take care of that.

jnweiger:
Yep, that strange scaling is specific to the local settings on this machine!
For normal installs, we'll keep the 1000 mm/s max.

PR #208, opened 2024-02-01
GenericGcodeDriver: Add support for grblHAL HTTP-upload. Fixes #207

### LINKED ISSUE #207: grblHAL http support
We'd like to use [grblHAL](https://github.com/grblHAL/) for our laser cutters so we started rewiring one to use the [picocnc](https://github.com/phil-barrett/PicoCNC) board with a pi pico. It would be nice if a file is first uploaded to an SD-card and then streamed from SD instead of streaming directly over the network (because of reliability). So we like to use the networking-plugin of grblhal and the http-server for uploading files.

The issue is, that the GenericGcodeDriver uses a different approach for http-upload (POST with X-filename header) than grblhal uses. See [here](https://esp3d.io/esp3d-webui/v3.x/documentation/api/fileupload/index.html), implemented [here](https://github.com/grblHAL/Plugin_networking/blob/master/http_upload.c#L97)

So I want to propose a setting for the GenericGcodeDriver modifying the behaviour of the http_upload-method. I'm willing to create a PR for that. Would this be acceptable? Or do you have other suggestions?

pimsierhuis:
See #207 

Used apache httpclient instead of native Java 11, because it was already used by the octoprint-implementation and the native client doesn't easily support multipart/form-data.

Decided not to refactor out the net.sf.corn-client for the current use in this pull request, because if that would break things it would be better handled separately. Can provide a separate PR for that.

### DIFF
--- a/src/main/java/de/thomas_oster/liblasercut/drivers/GenericGcodeDriver.java
+++ b/src/main/java/de/thomas_oster/liblasercut/drivers/GenericGcodeDriver.java
@@ -34,6 +34,7 @@ import de.thomas_oster.liblasercut.VectorPart;
 import de.thomas_oster.liblasercut.platform.Util;
 import net.sf.corn.httpclient.HttpClient;
 import net.sf.corn.httpclient.HttpResponse;
+import org.apache.http.client.methods.HttpGet;
 import purejavacomm.CommPort;
 import purejavacomm.CommPortIdentifier;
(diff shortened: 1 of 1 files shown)


pimsierhuis:
Please wait, I didn't test well enough. Seems there must be a websocket open before the /command http call works. In my tests it worked because I had webui running in parallel. Will implement the http_command with a websocket, that should do the trick I guess.

PR #220, opened 2024-04-03
GenericGcodeDriver: Add support for grblHAL HTTP-upload and static jo…

### LINKED ISSUE #207: grblHAL http support

pimsierhuis:
…b name. Fixes #207

### DIFF
--- a/src/main/java/de/thomas_oster/liblasercut/drivers/GenericGcodeDriver.java
+++ b/src/main/java/de/thomas_oster/liblasercut/drivers/GenericGcodeDriver.java
@@ -56,6 +56,8 @@ import java.net.Socket;
 import java.net.URI;
 import java.net.URISyntaxException;
 import java.nio.charset.StandardCharsets;
+import java.time.LocalDateTime;
+import java.time.format.DateTimeFormatter;
 import java.util.Enumeration;
 import java.util.LinkedList;
 import java.util.List;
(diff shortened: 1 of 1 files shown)


pimsierhuis:
I modified my previous pull request to a simpler variant where you can only upload a file and have the option to provide a static name. Autostart is more difficult to implement with grlbhal and not needed for our use case (we want to have a physical button on the machine). So if needed, maybe that can be implemented later.

Please merge this one :)

pimsierhuis:
@t-oster Do you have any chance to look at this? Sorry for being pushy by mentioning you, but I thought maybe you weren't noticed because the old PR was closed.

pimsierhuis:
Cool, thanks. If that week could be a tiny little bit shorter (like merging on monday so we could start using it tuesday), that would be great.. :)

PR #223, opened 2024-04-15
Support X and Y axis flip, mk. 2

TheAssassin:
Adds another commit to https://github.com/t-oster/LibLaserCut/pull/206 to fix the calculation of the bounding box.


mgmax:
Looks good, please let me know if everything works correctly (cut and engrave match, engrave image not mirrored, ...).

TheAssassin:
I tested all of this on an OMTech Polar. I could send pictures but I think they wouldn't prove anything (could be mirrored afterwards or so).

mgmax:
That's good. I just asked because recently we had some PRs by someone else where it turned out that the code doesn't work in real life.

